PR #151, opened 2014-05-12
Code style with autopep8 and initial steps for python 3 support

tbicr:
Look like code have many different code styles at once. I use auto tools as `autopep8` and manual checking for update code. Also it contain steps move to python 3 support.

Patches take care at most about code style and python 3 compatibility and didn't:
- line length, but now it can be too long
- double and single quotes for strings (for code used more single for docstring double)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -11,12 +11,13 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-import sys, os
+import sys
+import os
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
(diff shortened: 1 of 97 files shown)


jorisvandenbossche:
I would be very interested in having a python 3 compatible OWSlib. 
How is it going with this? 
Is there anything I can help with?

tomkralidis:
There has been some smaller PRs in the past months.  I suggest following this path.

jorisvandenbossche:
Ah, yes I see. The merged PRs are #225 and #226

tomkralidis:
@jorisvandenbossche to add, see related thread on owslib-devel: http://lists.osgeo.org/pipermail/owslib-devel/2015-April/000272.html

Issue #131, opened 2014-03-18
Support WaterML 2.0


rsignell-usgs:
Excerpted from the Ulmo package issue by @jirikadlec2
https://github.com/twdb/ulmo/issues/77,
but would be great if supported by OWSLib:

"It would be nice to support WaterML 2.0 in ulmo. WaterML 2 is a xml format for hydrologic data and it is an official standard of the OGC:
http://www.opengeospatial.org/standards/waterml

I think it's OK if only the getValues() method with an url parameter is supported.

An example WaterML 2.0 request is here:
http://data.fmi.fi/fmi-apikey/1cb8a6f3-d8f4-4ae7-8fc3-0b9449683cfb/wfs?request=getFeature&storedquery_id=fmi::observations::weather::timevaluepair&fmisid=101152&starttime=2013-08-01T00:00:00Z&endtime=2013-08-05T00:00:00Z&timestep=60&parameters=temperature

(this is a time-series of air temperature at a station in Finland)

mullenkamp:
Hi, has @peterataylor 's work been integrated?

peterataylor:
Hi @mullenkamp - yes it got merged in a while back by @kwilcox, see https://github.com/geopython/OWSLib/commit/c73c06eddd059836fcc7f0c2301c9becf8cf1897.

mullenkamp:
Thanks. I was only asking because this issue is still opened, but (as you said) it looked like it got implemented.

cehbrecht:
If I understand correctly this issue is closed by PR #241 since version 0.12.0. I close it now.

PR #171, opened 2014-08-05
SOS 2.0 and WaterML2.0 decoder

peterataylor:
Hi Tom,

As discussed in https://github.com/geopython/OWSLib/issues/131. 

Adding a decoder for SOS2.0 GetObservation requests that marshals XML responses from O&M and WaterML2.0 formats. 

Summarised again here:

* **[om.py](https://github.com/peterataylor/OWSLib/blob/master/owslib/swe/observation/om.py)** - handles O&M types. New classes: 
 * OM_Observation - handles base level O&M observation properties, with a result property that will vary depending on the type of observation
 * OM_MeasurementObservation - specialised OM_Observation that has a Measurement result (value + uom) 

* **[waterml2.py](https://github.com/peterataylor/OWSLib/blob/master/owslib/swe/observation/waterml2.py)** - handles WaterML2.0 types. New classes:
 * Timeseries, MeasurementTimeseries, TimeValuePair, TVPMetadata, TVPMeasurementMetadata, MeasurementTimeseriesDomainRange - types to support timeseries. Not all implemented at this stage, but enough to parse results from SOS2.0 instances, specifically the 52North implementations. 
  * MeasurementTimeseriesObservation - specialised OM_Observation that has a time-series as a result. 

* **[sos200.py](https://github.com/peterataylor/OWSLib/blob/master/owslib/swe/observation/sos200.py)** - added:
 * SOSGetObservationResponse - container for OM_Observation objects that result from a get_observation call to a SOS. 
 * ObservationDecoder - decoder class to handle different OM_Observation types. Currently supports OM_MeasurementObservation & MeasurementTimeseriesObservation. 

The [main test](https://github.com/peterataylor/OWSLib/blob/master/tests/doctests/sos_20_timeseries_decoder_ioos.txt) I've added shows basic usage.  

I've also changed the README file to be .MD rather than .txt so it renders on the landing page (https://github.com/peterataylor/OWSLib/commit/df622c707339f8b889553a217dcfd7909188f8c7, https://github.com/peterataylor/OWSLib/commit/48ea26b5f17354f9fa2447d8f1287ce5811b793d and https://github.com/peterataylor/OWSLib/commit/3730a677c7f322aa2f55af1dea7e594bfd308355). I can take these out of the PR if you prefer. 

Also added timeout to get_observation call (default 30 seconds) in sos200. Doctests set this to 120 as the services can be slow in responding. 

Updated the doco to include doc tests and supported versions: https://github.com/peterataylor/OWSLib/commit/3600cc830dec643bfb9f8762616231acbea59124.


kwilcox:
@peterataylor I'll look at this once the final python 3 compat PR is merged, thanks for the work

peterataylor:
Thanks Kyle. I've been meaning to get back to update this PR, but I'll wait for the v3 changes to go through. Let me know if/when you need a hand.

kwilcox:
See https://github.com/geopython/OWSLib/pull/241

PR #213, opened 2015-01-01
Use 'in' operator instead of has_key.

QuLogic:
It's not available in Python 3 and `in` is recommended.


QuLogic:
Travis failure appears to be due to external resources...

PS, does OWSLib really claim to support Python 2.5? I see mention of it in places, but then again the `tox.ini` does not have an env for it.

tomkralidis:
Thanks @QuLogic.  Maybe we should remove any refs to 2.5 in the codebase to be safe?

PR #214, opened 2015-01-01
Use absolute imports from __future__.

QuLogic:
Only the files where this makes a difference have been modified (for quicker review).

### DIFF
--- a/owslib/coverage/wcs110.py
+++ b/owslib/coverage/wcs110.py
@@ -11,7 +11,9 @@
 
 ##########NOTE: Does not conform to new interfaces yet #################
 
-from wcsBase import WCSBase, WCSCapabilitiesReader, ServiceException
+from __future__ import absolute_import
+
+from .wcsBase import WCSBase, WCSCapabilitiesReader, ServiceException
 from owslib.util import openURL, testXMLValue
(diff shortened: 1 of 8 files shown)


QuLogic:
I'm not sure if the build failures here are because of my changes, or just external services. I can't seem to reproduce the problems in `tox` though.

tomkralidis:
thanks @QuLogic for this PR.  I'm not very familiar using `__future__` or the implications of the changes here.

@kwilcox: comments/thoughts?

QuLogic:
The `absolute_import` setting simply means that modules within a package need to include the full path to their submodules (i.e., include the package name too), _or_ use a relative by prefixing with `.` (e.g., `from .etree import etree` in `owslib/wmc.py`). The "current directory" of a module is no longer searched in preference to the regular `sys.path`. It was added in Python 2.5.

tomkralidis:
@QuLogic thanks for this.  Do we need the `absolute_import` here as well, given we support Python 2.6+ ?

QuLogic:
`absolute_import` is 2.5+ and the default in 3.x. It makes a difference here since not all of your internal imports are absolute (i.e., `import owslib.<module>`).

PR #215, opened 2015-01-02
Fix two doctest failures.

QuLogic:
The first commit fixes a doctest that will fail on Python 3 because dictionaries are not guaranteed to be output in sorted order. But I've actually seen this problem with Python 2.7 in `tox`. So this changes the test to compare against the result, which I've seen used for all the other doctests that involve a dictionary.

The second commit fixes a doctest that is not guaranteed to be stable (since it depends on ordering from an external service).

With these two commits, I can get `tox` to pass locally except for the `tests/doctests/wps_response6.txt` doctest which times out.


QuLogic:
Test failure is external; maybe it should be disabled?

PR #217, opened 2015-01-05
Use new print function

QuLogic:
As with #214, for easier review, I have only added the `__future__` imports to files where the `print` statement/function changes would be made.

### DIFF
--- a/owslib/swe/common.py
+++ b/owslib/swe/common.py
@@ -1,3 +1,5 @@
+from __future__ import print_function
+
 from owslib.util import nspath_eval
 from owslib.namespaces import Namespaces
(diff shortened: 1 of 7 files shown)


QuLogic:
Test failed because of an external service's timeout; doesn't that get annoying?

tomkralidis:
@QuLogic thanks for this.  The `print` as a function works without the `from __future__ import print_function`.  Could we do without the imports?

QuLogic:
That is true, but only if `print` is passed a single argument (which is true in all cases here, actually). As it easy to forget, I still think adding the `__future__` import is useful.

And anyway, once we get through these need-to-review PRs, I was planning on opening a PR that adds all `__future__` imports to all files.

PR #219, opened 2015-01-07
Add all __future__ imports

QuLogic:
This PR adds all the (useful) `__future__` imports to all remaining files:
 * `absolute_import` was covered in #214 for files where it mattered. This PR just adds it to the remaining files.
 * `division` has no effect, since there are no divisions in the code. But it's future-proof to add it, and has no cost.
 * `print_function` was covered in #217 for files where it mattered. Again, this PR adds it to the remaining files.

There's also `unicode_literals` for full compatibility with Python 3. I did not add that one because it requires considerable care to ensure that you are not mixing `str` with `bytes` everywhere.

### DIFF
--- a/owslib/__init__.py
+++ b/owslib/__init__.py
@@ -1,1 +1,3 @@
+from __future__ import absolute_import
+
 __version__ = '0.8-dev'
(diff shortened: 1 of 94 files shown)


QuLogic:
Build failure is due to timeouts only.

PR #220, opened 2015-01-08
Fixes #218.

### LINKED ISSUE #218: Cannot create a collector using `set_datum` property.
Hi, I am experiencing an issue when creating a collector like this:

```python
from pyoos.collectors.coops.coops_sos import CoopsSos

collector = CoopsSos()
sos_name = 'water_surface_height_above_reference_datum'

datum = 'NAVD'
collector.set_datum(datum)
collector.end_time = stop
collector.start_time = start
collector.variables = [sos_name]
response = collector.raw(responseFormat="text/csv")
```

With OWSLib version > 0.8.9 I get `got an unexpected keyword 'result'`, but it works nicely with OWSLib <= 0.8.9.  The only change I see that could be the culprit is the passing of the keyword argument to openURL (that is absent in previous versions):

```python
response = openURL(base_url, data, method, username=self.username, password=self.password, **kwargs).read()
```

ocefpaf:
This PR should fix #218 (and https://github.com/ioos/pyoos/issues/43)

If I understood what @lukecampbell needed from the new keyword it should be `**url_kwargs` and not `**kwargs`.

### DIFF
--- a/owslib/swe/observation/sos100.py
+++ b/owslib/swe/observation/sos100.py
@@ -189,7 +189,8 @@ class SensorObservationService_1_0_0(object):
 
         data = urlencode(request)
 
-        response = openURL(base_url, data, method, username=self.username, password=self.password, **kwargs).read()
+        response = openURL(base_url, data, method, username=self.username,
+                           password=self.password, **url_kwargs).read()
         try:
             tr = etree.fromstring(response)
             if tr.tag == nspath_eval("ows:ExceptionReport", namespaces):


lukecampbell:
Yes this is correct.

lukecampbell:
This was to open up the possibility to specify a client-deemed timeout value to the URL request. Some SOS servers can have fairly large responses that take time to be compiled, transmitted and parsed which may exceed the timeout values of the request. It's not explicitly supported but this capability could be expanded in the future to support passing in additional HTTP parameters like authentication, encoding etc.

ocefpaf:
Nice idea BTW.

daf:
Wasn't @lukecampbell's changes exclusive to sos200.py?

ocefpaf:
@dap Nope, see:  https://github.com/geopython/OWSLib/commit/eafe603aa5a5cd677f0604836946884d85570ecf

ocefpaf:
Thanks @tomkralidis and sorry for the noise.  It took me a while to read the code properly and figure out that `url_kwargs != kwargs`.

tomkralidis:
No problem @ocefpaf and thanks for the fix!

lukecampbell:
Yes, thanks, I tested the first part but I never tested the second part, if I had, I would have observed the typo.

ocefpaf:
Not a problem, that is why we do open source development :wink:

PR #222, opened 2015-02-17
additions to SRV_ServiceIdentification

dblodgett-usgs:


owslib/iso.py
```diff
@@ -550,8 +553,19 @@ def __init__(self, md=None):
             self.operations = []
             self.operateson = []
         else:
+            self.title=None
```
tomkralidis: We don't need to init this here, since self.title below gets set no matter what

owslib/iso.py
```diff
@@ -550,8 +553,19 @@ def __init__(self, md=None):
...
             val=md.find(util.nspath_eval('gmd:citation/gmd:CI_Citation/gmd:title/gco:CharacterString', namespaces))
             self.title=util.testXMLValue(val)
+            
+            self.abstract=None
```
tomkralidis: We don't need to init this here, since `self.abstract` below gets set below no matter what via `util.testXMLValue`

tomkralidis:
@dblodgett-usgs thanks for this.  Overall comments (in addition to line comments above):

- can you squash this PR so we don't rewrite history?
- if possible, adding a test file / unit test would be great

dblodgett-usgs:
This this should do the trick. Sorry for the big mess of a PR before.

tomkralidis:
Perfect!  Thanks @dblodgett-usgs

PR #225, opened 2015-03-17
Some work on Python 3 compatibility

takluyver:
It seems that Python 3 compatibility has stalled (#151), and someone asked me to have a look at it.

Here's a new attempt, making mostly functional changes rather than trying to do style cleanup at the same time. It's not complete, but I thought this might be easier to review in smaller chunks. I used a mixture of python-modernize and manual changes to get to the point where I can run through the first few lines of the [CSW example](http://geopython.github.io/OWSLib/#csw). I initially started with the WMS example, but the URL used to demonstrate that is down.

### DIFF
--- a/owslib/csw.py
+++ b/owslib/csw.py
@@ -14,10 +14,18 @@ from __future__ import (absolute_import, division, print_function)
 import base64
 import inspect
 import warnings
-import StringIO
+import six
+try:
+    from StringIO import StringIO as BytesIO  # Python 2
+except ImportError:
+    from io import BytesIO  # Python 3
 import random
-from urllib import urlencode
-from urllib2 import Request, urlopen
+try:                    # Python 3
+    from urllib.request import urlopen, Request
+    from urllib.parse import urlencode
+except ImportError:     # Python 2
+    from urllib import urlencode
+    from urllib2 import Request, urlopen
 
 from owslib.util import OrderedDict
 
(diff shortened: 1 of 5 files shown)


owslib/csw.py
```diff
@@ -608,12 +616,17 @@ def _invoke(self):
                 if caller == 'getrecords2': caller = 'getrecords'
                 try:
                     op = self.get_operation_by_name(caller)
-                    post_verbs = filter(lambda x: x.get('type').lower() == 'post', op.methods)
+                    post_verbs = [x for x in op.methods if x.get('type').lower() == 'post']
                     if len(post_verbs) > 1:
                         # Filter by constraints.  We must match a PostEncoding of "XML"
-                        try:
-                            xml_post_url = next(x for x in filter(list, ([pv.get('url') for const in pv.get('constraints') if const.name.lower() == "postencoding" and 'xml' in map(lambda x: x.lower(), const.values)] for pv in post_verbs)))[0]
```
takluyver: This is one place where I did rewrite the code a fair bit. The enormous single line is very hard to follow, so I expanded it to a few lines of regular code - please check the logic carefully.

takluyver:
Ping @darkblue-b

darkblue-b:
hmm - will confer with colleagues - more soon

takluyver:
I guess you're happy with that approach. :-). I'll do some more of it.

PR #226, opened 2015-03-18
Further Python 3 compatibility work

takluyver:
The next chunk of compatibility work following #225.

### DIFF
--- a/owslib/swe/observation/sos100.py
+++ b/owslib/swe/observation/sos100.py
@@ -3,7 +3,10 @@ from __future__ import (absolute_import, division, print_function)
 import cgi
 from owslib.etree import etree
 from datetime import datetime
-from urllib import urlencode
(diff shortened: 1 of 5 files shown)


darkblue-b:
I checked out the Master; built and installed under Python 3.4.3; ran an interactive session in IPython 3.1.0-dev using the WMS protocol .. example PNG image and all object attributes succeeded.

tomkralidis:
cc @kwilcox 

@takluyver thanks much for these contributions.  I'm wondering whether we should turn on Python 3 testing in `.travis.yml` and / or update `tox.ini` ?

takluyver:
I ran the tests earlier, and there were still a bunch of failures on python
3, so you probably don't want to turn it on just yet. I will see how easy
it is to get things passing.

PR #227, opened 2015-03-21
Update wmts.py

darkblue-b:
copy py3 compat fix to wmts.py


tomkralidis:
@darkblue-b thanks.  Looks like a `NameError` at https://travis-ci.org/geopython/OWSLib/jobs/55317505#L676

kwilcox:
This was done in the py3 compat fixes: https://github.com/geopython/OWSLib/blob/master/owslib/wmts.py#L37-L42

PR #230, opened 2015-04-14
SOS namespace addition

gamesbook:
Adds the 'sa' namespace for handling SOS data.  An example from a GetObservation request:

```
<om:featureOfInterest xlink:title="my_sensor_id">
    <sa:SamplingPoint gml:id="my_sensor_id">
        <sa:sampledFeature xlink:href="urn:ogc:def:nil:OGC:unknown"/>
        <sa:position>
            <gml:Point gml:id="point_sf_42">
                <gml:pos srsName="urn:ogc:def:crs:EPSG::4326">15.751542 18.280503</gml:pos>
            </gml:Point>
        </sa:position>
    </sa:SamplingPoint>
</om:featureOfInterest>
```

### DIFF
--- a/owslib/namespaces.py
+++ b/owslib/namespaces.py
@@ -31,6 +31,7 @@ class Namespaces(object):
         'ows200':   'http://www.opengis.net/ows/2.0',
         'rim'   :   'urn:oasis:names:tc:ebxml-regrep:xsd:rim:3.0',
         'rdf'   :   'http://www.w3.org/1999/02/22-rdf-syntax-ns#',
+        'sa'    :   'http://www.opengis.net/sampling/1.0',
         'sml'   :   'http://www.opengis.net/sensorML/1.0.1',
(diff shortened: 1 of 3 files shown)


tomkralidis:
@gamesbook any chance we can add a test or two?

gamesbook:
Hi Tom

I have added a namespace test (but cannot get it to pass locally).  

I am not sure where to add a test for the GetObservation request patch?  If you can point me in the right direction, I will attempt to do so.

Derek

tomkralidis:
Hi Derek: thanks.  FYI I ended up merging this manually as there some merge issues and adding some SOS 2 compatibility.

To add a test case, you can either add a test file like https://github.com/geopython/OWSLib/blob/master/tests/doctests/sos_ngmp.txt or add to an existing test case.  Then feel free to issue another PR.

gamesbook:
Thanks Tom!  Sorry - I have been away for a month.

PR #233, opened 2015-04-29
Implemented getfeatureinfo function

granci:
Builds WMS getFeatureInfo request from given arguments. To prevent duplicity in the code this change added getmap_request function and modifies existing getmap function.

### DIFF
--- a/owslib/wms.py
+++ b/owslib/wms.py
@@ -165,15 +165,15 @@ class WebMapService(object):
             raise ServiceException(err_message, se_xml)
         return u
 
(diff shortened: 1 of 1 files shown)


tomkralidis:
@granci thanks.  Looks like there is an issue in one of the WMS tests: https://travis-ci.org/geopython/OWSLib/jobs/60523875#L374 (the non-WMS tests need to be addressed outside the scope of this PR).

kwilcox:
@granci Can you look at this test failure so this can be merged?

PR #236, opened 2015-05-08
Python 3 / requests support

### LINKED ISSUE #81: Python 3 support
Some of the changes are possible to do automatically with 2to3: http://pythonhosted.org/distribute/python3.html. I already tried that and manually tested some features. Some parts works but there is still code which doesn't work.

Blockers:
When running tests setup.py loads modules from source directory instead of build directory. Explained and fixed in: http://www.hackzine.org/python-testing-with-pytest-and-2to3-plus-tox-and-travis-ci.html. Anyone familiar enough with (2to3, )test.py and doctests to fix this?

### LINKED ISSUE #111: implement Python requests for HTTP
cc @kwilcox 

To track implementation of Python requests [1] for HTTP support across the codebase

Between GET / POST, authentication, cookies, etc., the requests library [1] would make OWSLib's HTTP request/response patterns minimal, solid, and very easy to 

Notes/ideas:
- perhaps a unified owslib.util.http_request function which all downstream code uses? Or just use directly?
- we should maintain http functions like `owslib.util.openURL`, `owslib.util.http_post` in `owslib.util`, as at least stubs (I've seen these in use in downstreams apps)
- functions like `owslib.util.bind_url` could probably be deleted afterwards

[1] http://docs.python-requests.org

daf:
This (rather large) PR updates OWSLib to support Python 3 (3.4 used in development) in addition to Python 2 - the codebase should work with both.

Started with [python-modernize](https://github.com/mitsuhiko/python-modernize), then continued manually.  @Jenselme contributed some very nice patches as well.  Requests was added to replace urllib/urllib2.

Fixes #111, fixes #81.

As per @tomkralidis comments in #111, `owslib.util.http_post` and `owslib.util.openURL` are retained, and a shim response object is used to maintain code compatibility.

Some rough edges/gotchas:
- Unicode/string handling changes cause trickiness, especially with doctests outputting strings.  In many cases I had to use `#doctest: +SKIP`.
- Exception reporting changed in Python 3, had to use `#doctest: +IGNORE_EXCEPTION_DETAIL` for them in doctests.  Not sure the extent of trouble there. 
- I'm not sure I've gotten the ResponseWrapper bit correct - I don't like having to encode the return value to utf-8, only to decode it again later (such as in WFS using StringIO).  I'm not sure what the best approach really is, or if this is just fine for now.

Please comment and/or suggest improvements, and I'll continue adding them to this branch, or you can open a PR against https://github.com/daf/OWSLib/tree/3.4


QuLogic:
__future__ imports only appear to be added where necessary. They should be added on all files (except maybe examples, not my decision, though) for consistency.

> Unicode/string handling changes cause trickiness, especially with doctests outputting strings. In many cases I had to use #doctest: +SKIP.

You could just wrap a great deal of them in `print()`. Or start using `future`, which prints strings consistently, but that's a bigger change.

daf:
> `__future__` imports only appear to be added where necessary. They should be added on all files (except maybe examples, not my decision, though) for consistency.

Sure, I'll add them.

> You could just wrap a great deal of them in `print()`. Or start using `future`, which prints strings consistently, but that's a bigger change.

I had enormous trouble using `print` for this being consistent across Python versions.  2.7 would complain like this:

```python
020     >>> print(wcs.provider.name)
Expected:
    Instituto Geográfico Nacional
Got:
    Instituto Geográfico Nacional
```

@QuLogic can you explain the second part of your statement, about "using `future`" - not sure what that means.

daf:
@QuLogic I only found one doctest file missing the `__future__` imports - you added the rest!

QuLogic:
> @QuLogic I only found one doctest file missing the __future__ imports - you added the rest!

How did I manage to find the only one with it!

> I had enormous trouble using print for this being consistent across Python versions.

Maybe try `# doctest: +WHITESPACE`?

> @QuLogic can you explain the second part of your statement, about "using future" - not sure what that means.

`future` is [a library for writing Python 3-but-compatible-with-2 code](http://python-future.org)

BTW, I would add `division` as well, even if it doesn't make much difference now.

tomkralidis:
@daf thanks!

QuLogic:
Whoa, just noticed you've got 3.2 builds too. I didn't think anyone attempted that much.

daf:
@QuLogic wasn't me, @Jenselme added in daf/OWSLib#2.  I had more trouble with 3.3!

PR #237, opened 2015-05-13
Return WMS layer names in the order defined in the WMS

geographika:
I'm currently working on a project where the order of WMS layers is important (as it is used for display order on a map).
The current wms.contents implementation uses an unordered dictionary. An OrderedDict should allow order to be retained and not require any other code changes. 

It does however require Python 2.7


geographika:
Just seen OrderedDict is used elsewhere.
Import should be ```from owslib.util import OrderedDict```

tomkralidis:
@geographika thanks for this.  Can we fix the [associated test failures](https://travis-ci.org/geopython/OWSLib/builds/62404557)?

geographika:
@tomkralidis I think several of the test failures are due to other issues. 
For example the csw_conterra test has HTTP errors as it tries to access https://gateway.ceh.ac.uk/soapServices/CSWStartup which no longer exists (the URL redirects). 

I have fixed 3 tests where the wms.contents checks for type dict rather than <class 'collections.OrderedDict'>

tomkralidis:
Thanks @geographika.  I can't seem to be able to merge.  Any chance you can rebase this against master or issue a fresh pull request?

geographika:
@tomkralidis I think the relevant tests had been fixed since my pull request. Other failing tests are due to old URLs.

tomkralidis:
@geographika I still had issues merging this.  I cleaned up your changes and merged in #259.  Let me know if something if off.  Closing.

geographika:
@tomkralidis - all looks good thanks! I had a merry dance of rebasing, merging, and playing with tox on Windows and ended up in a complete mess.

PR #241, opened 2015-05-22
SOS2.0 and WaterML2.0

kwilcox:
This is a manual rebase of https://github.com/geopython/OWSLib/pull/171 onto master along with a few small fixes in the PR.


tomkralidis:
@kwilcox thanks for this.  Looks like some [associated test failures](https://travis-ci.org/geopython/OWSLib/builds/63681696)?

kwilcox:
@tomkralidis all service endpoint test failures now

peterataylor:
Hi guys,

Any updates on this? The build failures were due to non responsive services called from the doc tests. Do you want to try and capture potential fails or change the service endpoint in the doctest? The general issue is service stability + longevity - can't really be sure all test services will be up.

kwilcox:
Slipped off my radar. I can rebase onto `master` so it can be merged.  The test services going down is a long standing issue... I just opened an issue for it: https://github.com/geopython/OWSLib/issues/312

peterataylor:
Thanks Kyle. That would be great if you could rebase; let me know if I can help.

kwilcox:
@peterataylor Do you know of any SOS2.0 servers I can rewrite the tests against? It took me so long... they have all gone away.

tomkralidis:
@kwilcox here's one: http://ngwd-bdnes.cits.nrcan.gc.ca/GinService/sos/gw

peterataylor:
Thanks @kwilcox . Actually I've recently been testing a new SOS2.0 service at the Bureau of Met here: http://www.bom.gov.au/waterdata/service. I have a py notebook using it here: http://nbviewer.jupyter.org/github/peterataylor/bom-water-data-service-example/blob/master/bom-water-sos-example.ipynb. 

This is likely to be quite a stable service - so a good candidate for the test suite. I can help to integrate it if need be.

domlowe:
Good idea Pete. Typo in the URL - should be:
http://www.bom.gov.au/waterdata/services

kwilcox:
@peterataylor I'm not having much luck with that service... does your notebook work? Do you have more commits to add?

peterataylor:
Hi @kwilcox - yes sorry there were two additional changes I had to make to get it fully working: https://github.com/peterataylor/OWSLib/commit/9274f8e1f06512d3ff2a7ed966d67b44f266309e. One is the way the Bureau service describes its endpoints in the caps document - the base URL is incorrect. I had to put in a temporary hack in to get it to work. I've reported the issue. We obviously don't want this fix in the library, so may be a blocker until we can get the service fixed..?

The second was adding observed_properties to the service metadata where the observed properties are not contained within an offering (they are inherited across all offerings).

kwilcox:
Please don't merge, waiting for other PRs to be merged first and ill rebase

kwilcox:
This is ready to merge. The test failures are all broken or changing external services (which we really need to fix).

owslib/iso.py
```diff
@@ -654,6 +654,9 @@ def __init__(self, md=None):
             val = md.find(util.nspath_eval('srv:accessProperties/gmd:MD_StandardOrderProcess/gmd:fees/gco:CharacterString', namespaces))
             self.fees = util.testXMLValue(val)
 
+            val = md.find(util.nspath_eval('srv:accessProperties/gmd:MD_StandardOrderProcess/gmd:fees/gco:CharacterString', namespaces))
```
tomkralidis: is this a duplicated by accident?

setup.py
```diff
@@ -14,7 +14,7 @@ def run_tests(self):
         errno = pytest.main(self.test_args)
         sys.exit(errno)
 
-readme = open('README.txt').read()
```
tomkralidis: Please keep README.txt as is as this will create packaging / manifest problems downstream.

owslib/iso.py
kwilcox: Yup bad merge, will remove

setup.py
kwilcox: I didn't even notice this!  I'll revert this to the .txt

tomkralidis:
Squashed and merged -- thanks @peterataylor / @kwilcox !

PR #260, opened 2015-07-03
Fix encoding problem for ResponseWrapper

drnextgis:
Using response encoding [here](https://github.com/geopython/OWSLib/blob/master/owslib/util.py#L133) is not reliable approach to get document encoding. Requests library only consider HTTP headers when attempting to work out text encodings and if `Content-Type` header is `text/xml` and it doesn't contain `charset` parameter then Request returns `ISO-8859-1` encoding:

```python
>>> from owslib.wms import WebMapService
>>> service = WebMapService('http://maps.kosmosnimki.ru/TileService.ashx/apikeyU96GP973UH', version='1.1.1')
>>> print service['7B3470B2C40E4B71B9AF9D1AFE17FA51'].abstract
ÐÑÑÐ¼ÑÐº
>>>
>>> import requests
>>> req = requests.get('http://maps.kosmosnimki.ru/TileService.ashx/apikeyU96GP973UH?SERVICE=WMS&Request=GetCapabilities&VERSION=1.1.1')
>>> req.encoding
'ISO-8859-1
>>> print req.content[:40]
<?xml version="1.0" encoding="utf-8"?>
```

### DIFF
--- a/tests/doctests/wfs_MapServerWFSCapabilities.txt
+++ b/tests/doctests/wfs_MapServerWFSCapabilities.txt
@@ -101,6 +101,6 @@ Test exceptions
 Lastly, test the getcapabilities method
 
     >>> wfs = WebFeatureService('http://nsidc.org/cgi-bin/atlas_south?', version='1.0')
-    >>> xml = wfs.getcapabilities().read().decode('utf-8')
-    >>> xml.find('<WFS_Capabilities') > 0
+    >>> xml = wfs.getcapabilities().read()
+    >>> xml.find(b'<WFS_Capabilities') > 0
     True
--- a/owslib/util.py
+++ b/owslib/util.py
@@ -129,10 +129,7 @@ class ResponseWrapper(object):
         return self._response.headers
 
     def read(self):
-        if not self._response.encoding:
-            return self._response.content           # bytes
-
-        return self._response.text.encode('utf-8')  # str
+        return self._response.content
 
     def geturl(self):
         return self._response.url
@@ -378,10 +375,7 @@ def http_post(url=None, request=None, lang='en-US', timeout=10, username=None, p
         rkwargs['auth'] = (username, password)
 
     up = requests.post(url, request, headers=headers, **rkwargs)
-    if not up.encoding:
-        return up.content           # bytes
-
-    return up.text.encode('utf-8')  # str
+    return up.content
 
 def element_to_string(element, encoding=None, xml_declaration=False):
     """


drnextgis:
@tomkralidis please review this PR.

kwilcox:
@drnextgis this caused a number of test failures, if you can review the failures and send some additional fixes I will review.  Thanks!

drnextgis:
@kwilcox I've fixed broken test. It seems that another test failures not caused by my PR.

kwilcox:
@QuLogic Any comments on this?  I don't use py3 and not sure of repercussions.

QuLogic:
If I understood the lxml FAQ correctly, you definitely _don't_ want to pass it Unicode, but bytestrings. So this change appears to make sense.

drnextgis:
Hi again, can you please review and merge if there no problems with it? Sorry to insist, but our system depends on OWSLib and we'd like to use upstream and not keep forks.

drnextgis:
I've squashed commits into one and made PR rebase. Is it any chance this PR will be merged? It looks like Travis CI is happy.

kwilcox:
I'll take a look first thing tomorrow!

drnextgis:
Awaiting your feedback. Thanks!

tests/doctests/wfs_MapServerWFSCapabilities.txt
```diff
@@ -101,6 +101,6 @@ Test exceptions
 Lastly, test the getcapabilities method
 
     >>> wfs = WebFeatureService('http://nsidc.org/cgi-bin/atlas_south?', version='1.0')
-    >>> xml = wfs.getcapabilities().read().decode('utf-8')
+    >>> xml = wfs.getcapabilities().read().decode('ISO-8859-1')
     >>> xml.find('<WFS_Capabilities') > 0
```
kwilcox: Why did this need to change?

drnextgis: If open GetCapabilities response you can see:

```
<?xml version='1.0' encoding="ISO-8859-1" ?>
<WFS_Capabilities ...
```
It means that Capabilities document encoding is `ISO-8859-1`. So for converting bytes to a string you have to specify document encoding:

```
xml = wfs.getcapabilities().read().decode('ISO-8859-1')
```

QuLogic: How about instead of decoding, just do `find` with a byte string?

drnextgis: In this case:

```
xml = wfs.getcapabilities().read()
xml.find('<WFS_Capabilities') > 0
```

you will get an error `TypeError("Type str doesn't support the buffer API",)` for Python3.

drnextgis: I've added `b` prefix and now this test works fine. Travis shows some fails but it is due to connection problems.

drnextgis:
@kwilcox I have updtaed test as suggested by @QuLogic. Travis CI is green.

PR #262, opened 2015-07-03
No need to explictly set "Content-length" header with requests

davidread:
We do not need to specify Content-length now as requests adds it in.

In fact it is helpful to remove it because there is a bug with older requests versions causing ConnectionError(!) - with this change it now works with requests 1.1. I know we specify >2.7 but all these little things help when dependency hell approaches...


davidread:
By the way, 'requests' fixed the issue in their release v1.2.3 (2013-05-25). But it would still be really helpful if this simple change was merged.

PR #263, opened 2015-07-03
Fix lxml import for ParseError, fix csw test

daf:
As pointed out by @drnextgis on 66200494ab59, I had the import for ParseError in the lxml path incorrect.  This patch fixes it, as well as an unrelated CSW test that had started failing because the source happened to be producing less records than when the test was written (moving targets and all).

### DIFF
--- a/owslib/etree.py
+++ b/owslib/etree.py
@@ -34,7 +34,8 @@ def patch_well_known_namespaces(etree_module):
 
 # try to find lxml or elementtree
 try:
-    from lxml import etree, ParseError
+    from lxml import etree
+    from lxml.etree import ParseError
     ElementType = etree._Element
 except ImportError:
     try:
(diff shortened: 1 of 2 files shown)


daf:
@tomkralidis the test failures on travis are only a timeout, same error on each one.

tomkralidis:
@daf what about this error? https://travis-ci.org/geopython/OWSLib/jobs/69449926#L386

daf:
Oh sure, make my spot checking look bad :)  I'll check it out..

daf:
And now a lesson in why you always check the easiest/most obvious thing first: NODC geoportal is returning 503s on every request. (this was after I had built a docker container for OWSLib to run python 3.2 as that was the easiest way to get that version going)

The error could likely be better (check HTTP code somewhere instead of trying to parse an error page) but that's a separate issue and not affected by this PR.

drnextgis:
@tomkralidis what reason this PR is still unmerged?

drnextgis:
Thanks!

PR #264, opened 2015-07-03
Reduce minimum 'requests' version

davidread:
Currently OWSlib requests a version of 'requests' which is only 2 months old. That is a very high bar when it is a library and most systems will not be as up to date. So I've tried lowering the minimum.

Tests pass with with this 2 yr old version of requests, aside from the 2 tests that reliably fail too with requests 2.7:
* tests/doctests/csw_ngdc.txt
* tests/doctests/ows_interfaces.txt

I've not done any other testing, so I guess merge it if you have confidence the tests have decent coverage.


davidread:
Tests pass with requests 1.0, now that #262 is passed. Only these two tests reliably fail, same as for requests 2.7:
* tests/doctests/csw_ngdc.txt
* tests/doctests/ows_interfaces.txt

kwilcox:
+1

tomkralidis:
@davidread I'm unable to merge this. Perhaps you can rebase? Not sure what the issue is.

davidread:
I've just pulled from master, so this should merge ok now.

tomkralidis:
THanks @davidread!

davidread:
Nice one, cheers

Issue #248, opened 2015-06-14
UnicodeEncodeErrors with repr/str


jorisvandenbossche:
Installed latest OWSlib 0.9.0, and noticed UnicodeEncodeErrors:

```
from owslib.sos import SensorObservationService
sos = SensorObservationService("http://sos.irceline.be/sos")
```

```
In [23]: sos.contents
...
UnicodeEncodeError: 'ascii' codec can't encode characters in position 49-50: ord
inal not in range(128)
```

But not sure this is an error in the code, or a problem in what the sos returns

tomkralidis:
Looks like it's tripping on this part of Capabilities:
```xml
<sos:ObservationOffering gml:id="89101 - PM1">
 <gml:name>Particulate Matter < 1 µm</gml:name>
...```

PR #265, opened 2015-07-03
Encode output of __str__ and __repr__ methods

drnextgis:
Encode string representation of objects. Fix issue #248.

### DIFF
--- a/owslib/ows.py
+++ b/owslib/ows.py
@@ -156,9 +156,9 @@ class Constraint(object):
 
     def __repr__(self):
         if self.values:
-            return "Constraint: %s - %s" % (self.name, self.values)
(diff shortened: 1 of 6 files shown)


tomkralidis:
@kwilcox any comments/feedback on this fix in general?  FYI we would need to check if more parts of the codebase can benefit:

```bash
grep -R "def __repr__" *
grep -R "def __str__" *
```

QuLogic:
This doesn't make sense on Python 3. The result should be a `str`, not `bytes`. Maybe what you want to do is apply the `python_2_unicode_compatible` decorator from `six`.

kwilcox:
:-1:

PR #267, opened 2015-07-19
Enable Travis' container-based infrastructure.

QuLogic:
Should start and run a little faster.


tomkralidis:
Ref: http://docs.travis-ci.com/user/workers/container-based-infrastructure/

PR #270, opened 2015-08-05
WFS 2.0 StoredQueries bugfixes/improvements

JuergenWeichand:
Repaired wfs.storedqueries (crash because of wrong timeout variable)
Repaired wfs.storedqueries (crash if storedquery has no abstract)
Added WFS 2.0 StoredQuery example
Added WFS 2.0 StoredQuery test


JuergenWeichand:
`tests/doctests/csw_gdp.txt` does fail because of HTTP 404. 
Issue is not in association with this PR.

PR #271, opened 2015-08-07
Add support for RESTful only WMTS

JuergenWeichand:
With this PR OWSLib will support **RESTful only WMTS**.
- will not crash if no operations are available in the capabilities
- added property function `.restonly`
- added function `.buildTileResource()` corresponding to `.buildTileRequest()`
- function `.gettile()` can now also be used for downloading a tile from a RESTful WMTS
- added test `wmts_RESTonly`
- updated test `wmts_geoserver21` (check if `restonly`)

### DIFF
--- a/owslib/wmts.py
+++ b/owslib/wmts.py
@@ -31,6 +31,7 @@ would be appreciated.
 
 from __future__ import (absolute_import, division, print_function)
 
+from random import randint
 import warnings
 import six
 from six.moves import filter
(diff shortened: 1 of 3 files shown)


JuergenWeichand:
Example for a RESTful only WMTSCapabilities (Mapproxy)
http://geoserv.weichand.de/mapproxy/wmts/1.0.0/WMTSCapabilities.xml

JuergenWeichand:
Some tests did fail because of connection problems (timeouts).

JuergenWeichand:
@tomkralidis thanks for the 0.9.1 release. What about this PR?

tomkralidis:
@JuergenWeichand thanks for the bump. Merged.

PR #272, opened 2015-08-08
WFS docs initial (WFS 1.1 and WFS 2.0)

JuergenWeichand:
Including a variety of examples.
* GetFeature using a `BoundingBox`
* GetFeature using a `Filter`
* GetFeature using `StoredQueries`


tomkralidis:
@JuergenWeichand thanks **very** much for the docs/contributions!!!

PR #275, opened 2015-08-25
wps: added ComplexDataInput class

cehbrecht:
Hi *,

i've added in wps the `ComplexDataInput` class. It can be used to add a complex input parameter (either an url reference or direct content). Before this there were only classes for special complex data input (FeatureCollection, ...).

Travis tests were all passing :)

Further changes:
* fix: unicode input is now possible for literal input data
* doctests for wps with `ComplexDataInput`: wps_request_7, ...
* doc: added note about testing and available anconda packages for owslib
* updated wps example wps-birdhouse.py
* moved the `IWebProcessingService` interface to interfaces.

Cheers,
Carsten


tomkralidis:
@cehbrecht thanks for the contribution.  At first glance it appears that Travis-ci tests fail for anything Python 3 (single failure at https://travis-ci.org/geopython/OWSLib/jobs/77119673#L1389).

cehbrecht:
sorry ... i've overseen that failure. I've fixed now the wps 52n test. Seems to be successful :)

Some owslib tests are "online" tests and they fail if a service is not available or too slow ...

tomkralidis:
Thanks @cehbrecht!

Issue #276, opened 2015-09-03
Issue a warning when not using lxml (and a new release of OWSLib)


ocefpaf:
It seems that not all the functionalities will work when `lxml` is not present.  It would be nice to issue a warning when OWSLib falls back to the default `etree`.  (See https://github.com/USGS-CIDA/pyGDP/issues/120)

PS: The bug fix https://github.com/geopython/OWSLib/commit/6f40078e0624079e266e00181b09e76c7956e733 is not in the 0.9.0 version.  We could use a new release of OWSLib :wink:

tomkralidis:
@ocefpaf thanks for the info.  OWSLib 0.9.1 released as per #278.

tomkralidis:
@ocefpaf I think this issue is downstream of OWSLib.  In your case, it is pyGDP which is depending on lxml semantics.  Issuing a warning in OWSLib would not be valid in this case, as OWSLib is guaranteed to work without lxml.

You can check which etree OWSLib is working with by a utlilty test like:

```python
from owslib.etree import etree

def which_etree():
        """decipher which etree library is being used by OWSLib"""
        which_etree = None

        if 'lxml' in etree.__file__:
            which_etree = 'lxml.etree'
        elif 'xml/etree' in etree.__file__:
            which_etree = 'xml.etree'
        elif 'elementree' in etree.__file__:
            which_etree = 'elementtree.ElementTree'

        return which_etree
```
Alternatively https://github.com/geopython/OWSLib/blob/master/owslib/etree.py can be update to include logging.  Pull requests welcome.

ocefpaf:
> @ocefpaf I think this issue is downstream of OWSLib.

Makes sense.  If `etree` not quacking like a `lxml` is not a problem for OWSLib, than the type checking should happen in pyGDP.

Thanks for the release.

tomkralidis:
@ocefpaf FYI I've since added this in #279 as a convenience function in master branch.

PR #279, opened 2015-09-08
add convenience function to report which etree OWSLib is using (#276)

tomkralidis:


ocefpaf:
Thanks!  I am forwarding this to the pyGDP devs.

PR #283, opened 2015-09-10
Add a children attribute to ContentMetadata to handle WMS nested layers

### LINKED ISSUE #8: getCapabilities with nested Layers
This getCapabilities File generated by ArcGIS Server 9.3.1 contains nested Layer definitions, where the non-queryable layers don't have a name.

owslib does not allow to get a valid list of layernames out of this getcapabilities response.

there are several issues: 1) the layer-reading should be fully recursive 2) if the layer has got no name defined, there is a problem in the dictionary, as the id is set to the name and thus the layer is not correctly inserted into the dictionnary.

Jenselme:
Inspired by the work of Ivan Boldyrev <lispnik@gmail.com> available
here: https://github.com/monoid/owslib/tree/tree

Close #8 

PS: I think it would be nice to add tests for this in OWSLib test suite. Do you know a public WMS with nested layers? I did my test on a private sever.

### DIFF
--- a/owslib/wms.py
+++ b/owslib/wms.py
@@ -123,13 +123,16 @@ class WebMapService(object):
         #recursively gather content metadata for all layer elements.
         #To the WebMapService.contents store only metadata of named layers.
         def gather_layers(parent_elem, parent_metadata):
+            layers = []
             for index, elem in enumerate(parent_elem.findall('Layer')):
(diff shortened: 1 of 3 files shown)


Jenselme:
The failing test is about TMS which I didn't modify.

Jenselme:
Just updated the PR with tests.

tomkralidis:
thanks @Jenselme!

Issue #298, opened 2015-12-14
ISO metadata gmd:LanguageCode


pmdias:
Hi,

Right now OWSlib, when parsing an ISO metadata xml file, uses the content of the element gmd:LanguageCode to get the languagecode, but probably should be using the content of the codeListValue attribute, as the ISO does not impose any restriction for the content of the element while the codeListValue is recommended to be from the ISO 639 codelist.

```python
val = md.find(util.nspath_eval('gmd:language/gmd:LanguageCode', namespaces))
self.languagecode = util.testXMLValue(val)
```

Using the codeListValue won't (almost certainly) break any current usage of OWSlib as almost everyone is putting the codeListValue also into the content of the element, but I believe that the correct behaviour, in this case, should be to use the codeListValue itself.

PR #299, opened 2015-12-14
md.languagecode value now comes from the codeListValue attribute of t…

pmdias:
This PR provides a solution to the issue raised at #298


tomkralidis:
Thanks @pmdias!

PR #305, opened 2015-12-28
Added support for spatialRepresentationType

pmdias:
Added the spatialRepresentationType element to the MD_Metadata.identification, supporting multiple instances of this element.

### DIFF
--- a/owslib/iso.py
+++ b/owslib/iso.py
@@ -303,6 +303,7 @@ class MD_DataIdentification(object):
             self.bbox = None
             self.temporalextent_start = None
             self.temporalextent_end = None
+            self.spatialrepresentationtype = []
         else:
             self.identtype = identtype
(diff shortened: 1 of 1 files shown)


owslib/iso.py
```diff
@@ -419,6 +420,12 @@ def __init__(self, md=None, identtype=None):
                 o = CI_ResponsibleParty(i)
                 self.contact.append(o)
 
+            self.spatialrepresentationtype = []
```
tomkralidis: Looks like this is def'd before as well?

pmdias:
Sorry @tomkralidis but I didn't understand what you meant with your question, maybe you could explain better?

tomkralidis:
Sorry @pmdias missed the broader context (on mobile). Looks good!

PR #315, opened 2016-03-15
disable ssl verification by default

cehbrecht:
i've added a flag to util.openURL() which disables ssl verification by default. I need this to work with wps services with self-signed certificates. 

Is there a better way to do this?

ssl verfication was enabled by default with python 2.7.9 (don't know the exact version).


kwilcox:
@cehbrecht Could you add this as an `__init__` parameter to the WPS classes you are using and pass it all the way through to `openURL()`? I'd rather have the default be the same as stable python and make the user disable it themselves. 

See https://github.com/geopython/OWSLib/pull/321 for a WMS example of doing something similar.

cehbrecht:
will open a new pull request for the changes.

Issue #313, opened 2016-03-08
WMS getmap passes kwargs improperly


pipes161:
The `getmap `function of `WebMapService `(line 201) improperly passes `kwargs `to `self.__build_getmap_request` (line 255).

`kwargs `should be passed as `**kwargs`, not `kwargs=kwargs`.

Version: 0.10.3

geographika:
I also ran into this today.

See pull request #316 otherwise extra params are sent to server as 'kwargs': {'MYPARAM': 1},

kwilcox:
Merged: https://github.com/geopython/OWSLib/pull/316

PR #316, opened 2016-03-15
Don't send kwargs as new dict

geographika:
To fix issue #313 otherwise extra params are sent to server as 'kwargs': {'MYPARAM': 1},


kwilcox:
Can you remove the `print()` statement?

PR #318, opened 2016-03-16
csw_linz.txt: doctest fix get more records.

micahcochran:
Fix to doctest csw_linz.txt to retrieve 20 records instead of 10.  The 'NZ Strata Parcels' data is no longer in the first 10 results.

This should fix the unit tests, which are currently failing.

Two of the tests failed.  The failure for Python 2.6 LXML=false, looks like the server aborted the connection.  The failure for Python 3.4 LXML=true, was a timeout error, perhaps Travis paused the process for more than 30 seconds (which could be increased, but should it?).


tomkralidis:
Thanks!

micahcochran:
Thank you.

PR #321, opened 2016-04-06
Support custom headers for WMS requests

drnextgis:
I've faced with the following issue. WMS server responds `403 Access denied` on all WMS requests with default [Requests](http://docs.python-requests.org/en/master/) `User-Agent` header:

```python
r = requests.get('http://maps.rosreestr.ru/arcgis/services/Cadastre/CadastreWMS/MapServer/WMSServer?SERVICE=WMS&VERSION=1.1.1&REQUEST=GetCapabilities')
r.text
u'<!DOCTYPE html>\n<html>\n  <head>\n    <title>403 Access denied. Code 0x10.</title>\n  </head>\n  <body>\n    <h1>Error 403 Access denied. Code 0x10.</h1>\n    <p>Access denied. Code 0x10.</p>\n    <h3>Guru Meditation:</h3>\n    <p>XID: 1053459697</p>\n    <hr>\n    <p>Varnish cache server</p>\n  </body>\n</html>\n'
```

But if set up custom header all works fine:

```python
>>> r = requests.get('http://maps.rosreestr.ru/arcgis/services/Cadastre/CadastreWMS/MapServer/WMSServer?SERVICE=WMS&VERSION=1.1.1&REQUEST=GetCapabilities', headers={'user-agent': 'OWSLib'})
>>> r.text
u'<?xml version="1.0" encoding="UTF-8"?>\n<WMT_MS_Capabilities version="1.1.1">...
```

This PR adds ability to use custom headers for WMS requests.

PR #326, opened 2016-05-02
added ssl verify parameter to util.openURL and wps

cehbrecht:
... ssl verification is enabled by default. When creating a wps one can disable the ssl verification:

    wps = WebProcessingService(url="https://locahost:8443/wps", verify=False)

In addition one can provide an optional "headers" parameter like in wms.

See #315 and #321.


cehbrecht:
tests all fail to due connection timeout in `tests/doctests/wmts_SFSWorldWMTSCapabilities.txt`

i've updated this patch according the comments in #315.

Do you know a better way to disable ssl verification?

cehbrecht:
this patch is now too old ...

PR #327, opened 2016-05-13
Fix wmts.txt test. Replace abstract, due change on the server's end. …

micahcochran:
… ~~Install libraries to silence SSL related InsecurePlatformWarnings in requirements-dev.txt.~~

This patch uses environment markers, which is a feature in pip version 6.0 (released 2014-12-22).

This test seems to be failing due to other tests failing.

**EDIT:** Separated SSL library installation related changes into PR #328

### DIFF
--- a/tests/doctests/wmts.txt
+++ b/tests/doctests/wmts.txt
@@ -14,8 +14,8 @@ Find out what a WMTS has to offer. Service metadata:
     '1.0.0'
     >>> wmts.identification.title
     'NASA Global Imagery Browse Services for EOSDIS'
-    >>> str.strip(wmts.identification.abstract)
-    'Near real time imagery from multiple NASA instruments'
+    >>> bytearray(wmts.identification.abstract, 'utf-8')
+    bytearray(b'The Global Imagery Browse Services (GIBS) system is a core EOSDIS component which provides a scalable, responsive, highly available, and community standards based set of imagery services.  These services are designed with the goal of advancing user interactions with EOSDIS\xe2\x80\x99 inter-disciplinary data through enhanced visual representation and discovery.')
     >>> wmts.identification.keywords
     ['World', 'Global']
 


requirements-2.6.txt
```diff
@@ -1 +0,0 @@
-ordereddict==1.1
```
tomkralidis: We need this file as `setup.py` uses it (for non-pip installs)

tests/doctests/wmts.txt
```diff
@@ -14,8 +14,8 @@ Find out what a WMTS has to offer. Service metadata:
     '1.0.0'
     >>> wmts.identification.title
     'NASA Global Imagery Browse Services for EOSDIS'
-    >>> str.strip(wmts.identification.abstract)
-    'Near real time imagery from multiple NASA instruments'
+    >>> str(wmts.identification.abstract.split('.')[0])
```
tomkralidis: Suggest we reference the full string to encapsulate the test properly.

tomkralidis:
Thanks for the contribution.  Suggest splitting this PR (wmts test fix, SSL libs, pip enhancements) given the various scope of the changes.

tests/doctests/wmts.txt
micahcochran: There is a Unicode character in the second part of the string.  I think it is an end single quote character.  I wasn't too sure how to make that work with Python 2 & 3, but I'll try again.

tomkralidis: ```u'The Global Imagery Browse Services (GIBS) system is a core EOSDIS component which provides a scalable, responsive, highly available, and community standards based set of imagery services.  These services are designed with the goal of advancing user interactions with EOSDIS\u2019 inter-disciplinary data through enhanced visual representation and discovery.'```

micahcochran:
OK. I'll make separate PRs for the changes.

requirements-2.6.txt
micahcochran: I'll entirely drop that change.

micahcochran:
I re-pushed the commit to only contain the fix for wmts.txt test.

tomkralidis:
Thanks @micahcochran!

micahcochran:
Thanks!

PR #331, opened 2016-05-18
Fix #129: Add username and password arguments to WFS class constructors

### LINKED ISSUE #129: Authentication on wfs
Maybe this is due to another issue (missing docs for the wfs).
But as far as I could see, there is no option for authentication in the wfs.

In wms you can add username and password in the init.

probably this will be fixed when this issue is rsolverd: https://github.com/geopython/OWSLib/issues/111

But this might take a while (?).

I Am doubting (because I need to use this in short term) to clone the repo and add a comparable solution like in wms...

Can you let me know this is a good idea?
or did I maybe oversee another easier way to do the logon on wfs?

BasicWolf:
This changeset allows passing low-level HTTP authentication details
to high level WebFeatureService API.

WFSCapabilitiesReader was moved to features.common in order to avoid duplication.

### DIFF
--- a/owslib/feature/common.py
+++ b/owslib/feature/common.py
@@ -1,0 +1,66 @@
+import cgi
+from owslib.etree import etree
+from owslib.util import openURL
+
+try:
+    from urllib import urlencode
+except ImportError:
+    from urllib.parse import urlencode
+
+
+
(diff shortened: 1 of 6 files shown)


kwilcox:
Thanks @BasicWolf!

PR #338, opened 2016-07-29
patched wms to read caps from ncwms2 with default namespace

cehbrecht:
… and timepositions.

ncWMS2 uses a xml default namespace which failed the GetCapabilitiesReader. The solution was to remove the default namespace:

http://stackoverflow.com/questions/34009992/python-elementtree-default-namespace#35165997

This patch also strips the timepositions to remove newlines.

This patch includes a doctest.

### DIFF
--- a/owslib/wms.py
+++ b/owslib/wms.py
@@ -26,6 +26,7 @@ except ImportError:     # Python 2
 import warnings
 
 import six
+import re
 
 from .etree import etree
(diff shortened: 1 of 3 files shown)


tomkralidis:
Shouldn't this be an ncWMS bug?  WMS 1.1.1 does not define a namespace.

cehbrecht:
it is fixed on the ncWMS side and available with the next release.

PR #340, opened 2016-08-02
added verify option to util.openURL

cehbrecht:
ssl verification is enabled by default and can be disabled with ``verify=False``. This flag can be used by wps, wms, etc ... when working with self-signed certificates.

http://docs.python-requests.org/en/latest/user/advanced/?highlight=ssl#ssl-cert-verification

### DIFF
--- a/owslib/util.py
+++ b/owslib/util.py
@@ -150,12 +150,15 @@ class ResponseWrapper(object):
     # @TODO: __getattribute__ for poking at response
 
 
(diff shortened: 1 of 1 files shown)


cehbrecht:
@tomkralidis 
Reviewed myself. Test are passing except for a timeout of a WMS service. It is only an optional parameter. The "real" changes need to happen in the WPS etc modules.

PR #341, opened 2016-08-30
Changing conda install instructions

rsignell-usgs:
Most everyone will want to use conda-forge channel for conda install, so making that more clear.  
Nearly all the packages that used to be on the ioos channel are now on conda-forge

### DIFF
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -97,13 +97,13 @@ Anaconda:
 
 .. note::
 
-   The OWSLib conda packages are **not official** and provided on different conda channels: http://anaconda.org/search?q=type%3Aconda+owslib
+   The OWSLib conda packages are provided by the community, not OSGEO, and therefore there may be 
+   multiple packages available.  To search all conda channels: http://anaconda.org/search?q=type%3Aconda+owslib
+   However usually conda-forge will be the most up-to-date. 
 
 .. code-block:: bash
 
-  $ conda install -c birdhouse owslib
-  # or
-  $ conda install -c ioos owslib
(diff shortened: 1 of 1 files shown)


rsignell-usgs:
@ocefpaf , please check this proposed change

docs/en/index.rst
```diff
@@ -97,13 +97,13 @@ Anaconda:
...
+   The OWSLib conda packages are provided by the community, not OSGEO, and therefore there may be 
+   multiple packages available.  To search all conda channels: http://anaconda.org/search?q=type%3Aconda+owslib
+   However usually conda-forge will be the most up-to-date. 
 
 .. code-block:: bash
 
-  $ conda install -c birdhouse owslib
-  # or
-  $ conda install -c ioos owslib
+  $ conda install -c conda-forge owslib
```
ocefpaf: Looks good to me. :+1: 

@tomkralidis if you want to become a maintainer for `owslib` in `conda-forge` let me know.
All that is needed is a PR adding your GitHub handle [here](https://github.com/conda-forge/owslib-feedstock/blob/master/recipe/meta.yaml#L43-L46) and, once merged, you will be automatically added to the feedstock (the conda recipe repository) team.

tomkralidis: cc @kalxas.

@ocefpaf sounds good.  Note that @kalxas is better suited here I would recommend him instead if he's interested.

rsignell-usgs:
Does something need to be run to update the page here?
http://geopython.github.io/OWSLib/#installation

tomkralidis:
The webpage will be updated on next release.

PR #344, opened 2016-09-20
Add possibility to add authentication token has headers for a WebProcessingService invocation. 

mmelcot:
Example: auth_headers = {'Access_token': auth_token, 'Authorization': auth_token}

### DIFF
--- a/owslib/wps.py
+++ b/owslib/wps.py
@@ -180,7 +180,7 @@ class WebProcessingService(object):
     Implements IWebProcessingService.
     """
 
-    def __init__(self, url, version=WPS_DEFAULT_VERSION, username=None, password=None, verbose=False, skip_caps=False):
+    def __init__(self, url, version=WPS_DEFAULT_VERSION, username=None, password=None, verbose=False, skip_caps=False, auth_token=None, headers=None):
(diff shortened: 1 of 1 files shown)


cehbrecht:
@mmelcot 
just saw your patch and i'm curious to know what kind of auth_tokens you use. We also need to protect our wps services and want to use simple auth-tokens. We have a prototype of an ows security proxy called twitcher:

https://github.com/bird-house/twitcher

Auth tokens are currently just uuid strings. They can be provided either as part of the request url or as a header variable. With `owslib` i'm using the `headers` parameter ... well, it's not merged yet ... 

https://github.com/bird-house/OWSLib/blob/pingudev/owslib/wps.py#L184

The `headers` parameter is also used in `owslib.wms`.

mmelcot:
@cehbrecht
We use token from an [OpenAM](https://forgerock.org/openam/) server.

We didn't set-up the OpenAM so I can't tell you how complex it is. But on the client side it is very simple since you just have to provide headers like this:
headers = {'X-OpenAM-Username': login, 'X-OpenAM-Password': password}
in order to receive a token.

PR #350, opened 2016-12-09
Fix WFS2.GetFeature parameter names

arnaud-morvan:
Some WFS GetFeature parameter names has changed with WFS2.
* ``typename`` => ``typenames``
* ``maxfeatures`` => ``count``

### DIFF
--- a/owslib/feature/__init__.py
+++ b/owslib/feature/__init__.py
@@ -128,13 +128,19 @@ class WebFeatureService_(object):
             request['query'] = str(filter)
         if typename:
             typename = [typename] if type(typename) == type("") else typename
-            request['typename'] = ','.join(typename)
+            if int(self.version.split('.')[0]) >= 2:
+                request['typenames'] = ','.join(typename)
+            else:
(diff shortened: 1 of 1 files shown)


tomkralidis:
Should we do the version check and update of params once and at the end of codepath?  Example:

```
if int(self.version.split('.')[0]) >= 2: 
    if 'typename' in request:
        request['typenames'] = request.pop('typename')
    if 'maxfeatures' in request:
        request['count'] = request.pop('maxfeatures')
```

Not necessarily less code, but would this be cleaner?  Not sure.

arnaud-morvan:
No opinion on this.

My hesitation was to add (or not) "count" and "typenames" parameters to getGETGetFeatureRequest with something like this : 
```
if maxfeatures or count: 
    if int(self.version.split('.')[0]) >= 2:
        request['count'] = str(count) or str(maxfeatures)
    else:
        request['maxfeatures'] = str(maxfeatures) or str(count)
```
Because ``maxfeatures`` is useful for old projects compatibility, but not really clear for new project using WFS2 only.

PR #352, opened 2016-12-21
Fix #351: Remove srsname validation in WFS 1.1.0 getfeature routine

### LINKED ISSUE #351: srsname override in WFS 1.1.0
OWSLib allows specifying `srsname` argument in WFS GetFeature requests. However in WFS 1.1.0 module the value passes validation, which denies supplying arbitrary values. This validation is rather restrictive and would not allow e.g. `'EPSG:3301'` value, requring `'urn:x-ogc:def:crs:EPSG:3301'`. 

I strongly believe this validation should be removed, because the explicit `srsname` format in some cases affects the output. For example, GeoServer [axis ordering](http://docs.geoserver.org/latest/en/user/services/wfs/basics.html#wfs-basics-axis) of coordinates depends on whether a short or a full representation of srsname is used.

BasicWolf:


drnextgis:
Faced with the same issue. The list of SRSs from capabilities document does not limit the actual target SRS usage in GetFeature requests [[1]](http://docs.geoserver.org/stable/en/user/services/wfs/webadmin.html#configuring-additional-srs). @tomkralidis could you consider merging this PR?

tomkralidis:
@drnextgis can you provide a test case/snippet of where this is problematic?  I'm also wondering if we should also throw a warning.

drnextgis:
For example:

```python

wfs = WebFeatureService(url='http://lipas.cc.jyu.fi/geoserver/lipas/wfs',
                        version='1.1.0')
response = wfs.getfeature(
    typename='lipas:lipas_5370_jaaspeedway_rata',
    maxfeatures=1,
    srsname='EPSG:3857')
```

But now OWSLib throws an error:

```python
  File "<stdin>", line 4, in <module>
  File "/home/dr/ngw/env/local/lib/python2.7/site-packages/owslib/feature/wfs110.py", line 207, in getfeature
    raise ServiceException("SRSNAME %s not supported.  Options: %s" % (srsname, options))
owslib.util.ServiceException: SRSNAME EPSG:3857 not supported.  Options: urn:x-ogc:def:crs:EPSG:3067
```

But if you try send the same request manually - all works fine: http://lipas.cc.jyu.fi/geoserver/lipas/wfs?srsname=EPSG:3857&service=WFS&maxfeatures=1&request=GetFeature&typename=lipas:lipas_5370_jaaspeedway_rata&version=1.1.0

I've made a more general PR related to this issue: https://github.com/geopython/OWSLib/pull/405

drnextgis:
Issue was fixed. This PR can be closed.

PR #353, opened 2016-12-22
Fixing doc in CSW-T - missing "" in parameter

pmauduit:
Minor modification: missing quote between argument in open() in doc


tomkralidis:
Thanks @pmauduit!

PR #358, opened 2017-01-13
Add doseq to WFS request query urlencode

latenssi:
Allow repetitive query parameters in WFS request

### DIFF
--- a/owslib/feature/__init__.py
+++ b/owslib/feature/__init__.py
@@ -150,7 +150,7 @@ class WebFeatureService_(object):
         if outputFormat is not None:
             request["outputFormat"] = outputFormat
 
-        data = urlencode(request)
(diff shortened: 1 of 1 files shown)


tomkralidis:
@latenssi any example of a user case of where this is required?

latenssi:
@tomkralidis We are accessing an OWS compatible endpoint. In this endpoint it is possible to send repetitive 'latlon' or 'place' query parameters to get a response which contains values for each. 

I am not sure if this is a standard but since it was already used [here](https://github.com/geopython/OWSLib/blob/2f187d7afa6f306788004cfa9f63d97d0475f712/owslib/wmts.py#L791) I thought it would be ok (and it helps us a lot)

tomkralidis:
Thanks @latenssi for the clarification.

PR #359, opened 2017-01-18
Swe fixes

benjwadams:
Fixes an issue with optional SWE elements where optional time element would raise an uncaught exception if the element was not found.

Removed use of eval to delegate to class constructors, and instead used a dict generated from the classes contained within the module, mapping the string name as a key to the corresponding class.

Code in question is here:
https://github.com/geopython/OWSLib/blob/c86c02f/owslib/swe/common.py#L67-L72

The eval could previously run arbitrary code conforming to valid XML tag names with a properly crafted XML payload.

For example, passing the etree parsed element to `NamedObject`'s constructor with contents of:

"<root name='foo'><etree.tostring></etree.tostring></root>"

is valid payload and will result in a string of '<etree.tostring/>' being returned, which is not desirable.

Fortunately, restrictions on XML tag characters prevent anything too malicious from happening in the current codebase, i.e. no parens or quotes so you can't do `__import__('os').system('rm -rf /important_dir')`

### DIFF
--- a/owslib/swe/common.py
+++ b/owslib/swe/common.py
@@ -268,6 +268,10 @@ def get_time(value, referenceTime, uom):
         elif value.lower() == "-inf":
             value  = NegativeInfiniteDateTime()
 
+    # Usually due to not finding the element
+    except TypeError:
+        value = None
+
     return value
 
 class Time(AbstractSimpleComponent):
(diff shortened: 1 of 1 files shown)


owslib/swe/common.py
```diff
@@ -17,6 +20,7 @@ def get_namespaces():
 def nspv(path):
     return nspath_eval(path, namespaces)
 
+from sys import modules
```
tomkralidis: This line is specified above already.

daf:
:+1:

PR #372, opened 2017-06-06
fix #248

### LINKED ISSUE #248: UnicodeEncodeErrors with repr/str

pesekon2:
unicode encodings in SOS GetCapabilities

### DIFF
--- a/owslib/swe/observation/sos100.py
+++ b/owslib/swe/observation/sos100.py
@@ -1,6 +1,7 @@
 from __future__ import (absolute_import, division, print_function)
 
 import cgi
+import sys
 from owslib.etree import etree
 from datetime import datetime
 try:                    # Python 3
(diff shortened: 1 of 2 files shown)


kwilcox:
Did you try to use [this six method](https://pythonhosted.org/six/#six.python_2_unicode_compatible) before reverting to the conditionals?

pesekon2:
I did, but it didn't worked for me. Now I have found why: https://github.com/benjaminp/six/issues/148

It doesn't support __repr__. I think it should be OK to use this reverting and recode it after the __repr__ support. But if you wish, I can try to use six and rewrite somehow the __repr__ method.

pesekon2:
Hi again. So what do you think about it?

cehbrecht:
@pesekon2 
I close the PR in favor of #477. Improvements can be added in follow-up pull-requests.

PR #373, opened 2017-06-26
added proxies feature to openURL, fixes #371

### LINKED ISSUE #371: Feature: openURL using a proxy?
One of our next measurement campaigns will be in a region with a very small network bandwidth. 
We are thinking about using a local proxy to cache all images requests from our servers.
Have you thought on adding a proxy parameter? Or should we do this on our library by a prepared request?

ReimarBauer:


tomkralidis:
@ReimarBauer thanks for this contribution.  Can you provide examples/update docs on how one can use/call/pass this new parameter?

ReimarBauer:
Hi @tomkralidis 

we use a squid server which proxies for http and https.  This gives a speedup in regions where your internal network is much faster then the internet.

proxies = {
    "http": "http://yoursquidproxy:3128",
    "https": "http://yoursquidproxy:3128"
  }

u = openURL(base_url, data, method, username=username, password=password, proxies=proxies)

Please point me to the documentation system to add an example.

tomkralidis:
@ReimarBauer can you add an example workflow in https://github.com/geopython/OWSLib/blob/master/docs/en/index.rst by adding a section and showing how to connect to a service with a proxy.  Thanks.

cehbrecht:
The current conflict in `owslib/util.py` is easy to solve and the PR just adds a `proxies` parameter which is delegated to `requests`, like `headers` etc. 

Still I hesitate to merge this PR because the `requests` documentation says one can also use environment variables (`HTTP_PROXY`) to set a proxy:
http://docs.python-requests.org/en/latest/user/advanced/#proxies

@ReimarBauer I will close this PR for now. If you think it is worth to have the `proxies` parameter then please provide a new PR including a description. Thanks.

ReimarBauer:
I was not aware that you want a copy of the request documentation.  But anyway I have also not readed the HTTP_PROXY environment in the docs, I will try this or keep my version.

cehbrecht:
Would be nice if the `HTTP_PROXY` environment works and it could be mentioned in the docs. If you need to set it in `openURL` and want to use it in the modules `wms`, `wps`, ... then you have to pass down this parameter which is a nasty thing. I did this already for some parameters used by `wps`, see PR 438.

PR #383, opened 2017-08-28
 - fixes #382: WPS DescribeProcess issue on DataType

### LINKED ISSUE #382: WPS DescribeProcess issue on DataType
Currently WPS DescribeProcess works only for DataType elements declaring the reference as attribute

e.g.:

`<ows:DataType ows:reference="xs:anyURI" xmlns:ows="http://www.opengis.net/ows/1.1" />`

The code assumes this as the unique valid way to declare them and blindly parses the reference as:

`sub_element.get(nspath("reference", ns=subns)).split(':')[-1]`

resulting, of course, on a unmanaged Exception in the case `sub_element.get(nspath("reference", ns=subns))` is `None`

Moreover for the WPS Protocol is legit to declare DataType as follows:

`<ows:DataType xmlns:ows="http://www.opengis.net/ows/1.1">xs:anyURI</ows:DataType>`

My suggestion is to update the OWSLib[wps.py 920:928] like below:

```
        if literal_data_element is not None:
            self.dataType = 'LiteralData'
            for sub_element in literal_data_element:
                subns = getNamespace(sub_element)
                if sub_element.tag.endswith('DataType'):
                    reference = sub_element.get(nspath("reference", ns=subns)) or sub_element.text
                    if reference and ':' in reference:
                        self.dataType = reference.split(':')[-1]
```

afabiani:


owslib/__init__.py
```diff
@@ -1,3 +1,3 @@
 from __future__ import (absolute_import, division, print_function)
 
-__version__ = '0.14-dev'
+__version__ = '0.14-geonode-dev'
```
tomkralidis: Can you remove this change?

afabiani:
@tomkralidis reverted inii.py version.

Issue #379, opened 2017-08-03
Feature: WPS Request with Bounding Box


tgmendes:
Hello,

We are using OWSLib to make WPS requests, and we found out that there is no clear support to send requests with a defined bounding box (only the ability to parse them). Is this true?

If so, is there a workaround or is there a plan to include this feature?

Regards,
Tiago

cehbrecht:
this is also on my wish list to make the bbox work. I did already some work on my fork:
https://github.com/bird-house/OWSLib/blob/pingudev/owslib/wps.py#L1449

But it is not finished ... i can just remember that it didn't work yet with pywps-4.

tgmendes:
Thank you very much!

Your implementation worked for us, we just had to make some minor modifications to adapt to our needs.

Cheers!

cehbrecht:
Thanks for testing :) 

My patched owslib fork is also available on a conda channel:
https://anaconda.org/birdhouse/owslib/files

But i will check this bbox code and try to get it into the official owslib.

cehbrecht:
@tgmendes 
... forgot to ask: do you have your modifications on a github fork? Which WPS service are you using?

tgmendes:
We did not place any modifications on a github fork because we only created a wrapper so that the BBOX parameters were passed as individual parameters instead of a string or list. If you find this interesting I can send you the file or place it in a fork.

Regarding WPS, we are using a private WPS service, which is maintained by Terradue.

cehbrecht:
@tgmendes 
if you don't mind you could add a gist to this ticket (https://gist.github.com/). Thanks.

tgmendes:
Here you go:
https://gist.github.com/tgmendes/58d0767ca1a5950c2f91d661912202ac

cehbrecht:
This is fixed with PR #384.

PR #384, opened 2017-09-04
New feature: wps request with bbox (#379)

cehbrecht:
This PR implement feature request #379, sending a `BoundingBoxDataInput` parameter with a WPS request. It fixes also the parsing of a BoundingBox XML element.

A test for the usage of the new `BoundingBoxDataInput` has been added.


cehbrecht:
all WPS tests pass ... the other test failures are not caused by this patch.

PR #396, opened 2017-11-28
Fix getTypedValue for booleans.

Roel:
xs:boolean is either 'true' or 'false', but since bool('false') returns
True, we better check if the string equals 'true'.


johanvdw:
Roel,
consider using https://docs.python.org/2/distutils/apiref.html?highlight=distutils.util#distutils.util.strtobool

johanvdw:
My old comment was wrong  (as xsd forces either true or false). This should be merged - it is wrong in the old code.

PR #397, opened 2017-11-29
Add parsing of FC_FeatureCatalogue to CSW

Roel:
I tried adding the parsing of gfc:FC_FeatureCatalogue's to the csw module, allowing to get these results from a csw service. Feature catalogues offer more information of the feature type of a vector layer, containing f.ex. a full description of each attribute or the domain (possible values) of an attribute with a description.

I added the parsing of gfc:FC_FeatureCatalogue under the 'gmd' outputSchema, since this is how Geonetwork treats them as well.

I also added the gmd:contentInfo (containing gmd:MD_FeatureCatalogueDescription) to the MD_Metadata class, because this is the way the metadata of a layer is linked with the feature catalogue.

I included the (in my opinion) most useful fields in each type. In most cases, the schema allows more fields to be present but these are not parsed. There is always the 'xml' field containing the raw content for users to parse it themselves.

I'm happy to answer questions or discuss improvements!

### DIFF
--- a/owslib/csw.py
+++ b/owslib/csw.py
@@ -30,7 +30,7 @@ from owslib.etree import etree
 from owslib import fes
 from owslib import util
 from owslib import ows
-from owslib.iso import MD_Metadata
+from owslib.iso import MD_Metadata, FC_FeatureCatalogue
 from owslib.fgdc import Metadata
 from owslib.dif import DIF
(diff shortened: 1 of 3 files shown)


tomkralidis:
Thanks, awesome feature!

PR #399, opened 2017-12-01
Improve remote metadata parsing.

Roel:
I made some improvements to the remote metadata parsing of WMS and WFS layers. I noticed some of the fields of MetadataURL elements (like the format or the type) were incorrectly parsed (especially for WFS), so I updated these based on the respective XSD schemas.

I also tried making the parsing of remote metadata based on these MetadataURL's more robust (f.ex. by looking for gmd:MD_Metadata elements in the resulting tree instead of taking the root element, which is often not gmd:MD_Metadata but a csw:getRecordByIdResponse).

Thirdly I moved the parsing of remote metadata into a seperate method. This gives control of this expensive operation to the user, allowing f.ex. parsing only the remote metadata of certain layers of interest. I kept the 'parse_remote_metadata' argument for the services for compatibility (and this should still work).

Lastly I added a helper method to more easily retrieve a list of all parsed remote metadata documents for a layer.

To sum it up, with this PR we allow flows like:
```
wfs = WebFeatureService(..., parse_remote_metadata=False)
for interesting_layer in ['interesting:layer1', 'interesting:layer2']:
    layer = wfs.contents[interesting_layer]
    layer.parse_remote_metadata()
    mdrecords = layer.get_metadata()
    for m in mdrecords:
        print(m.identification.title)
```

Off course, I'm happy to discuss improvements!


johanvdw:
@tomkralidis : I reviewed this pull request and it looks okay to me. As far as I can tell your concern that there was already some metadata parsing in wms is adressed as well (common abstract class - similar implementation).

johanvdw:
Note I used this example to test - it should be easy to convert it into a unit test, but I think we should merge master first for that  (new test style).
```

wfs = WebFeatureService(url="https://www.dov.vlaanderen.be/geoserver/wfs", version="2.0.0")

layer = wfs.contents['dov-pub:Boringen']
layer.parse_remote_metadata()
mdrecords = layer.get_metadata()
for m in mdrecords:
    print(m.identifier)
    print(m.identification.title)
```

cehbrecht:
@Roel Do you like to add a test like @johanvdw proposed to this PR? You need to merge with master first.

Roel:
Sure, I'll try to add it this weekend!

Roel:
Can I use a remote webservice in the test or do you prefer local testdata?

cehbrecht:
@Roel For the test-suite it would be better to have offline tests. Faster and less likely to break :) But of course you can also add an "online" test with a remote service. Please use the pytest markers in that case (`mark.online`, `mark.skipif`):
https://github.com/geopython/OWSLib/blob/723037958906262252f02452b8248939ab939d84/tests/test_wms_getmap.py#L14

Some owslib modules allow to use a prepared service response ... I'm sure for `owslib.wps`, but might be available for others as well. This would be kind of a **mock** test:
https://github.com/geopython/OWSLib/blob/723037958906262252f02452b8248939ab939d84/tests/test_wps_describeprocess_bbox.py#L11

If you like, you may add both an online and offline test.

To make offline tests more reliable it would be nice if we would have our own test-services we could control. @tomkralidis Tom, are you aware of online OWS test-services?

Roel:
Thanks for the info!

In another project (which uses OWSLib, in fact), I use PyTest's [monkeypatch fixture](https://docs.pytest.org/en/latest/monkeypatch.html) to mock network calls:
https://github.com/DOV-Vlaanderen/pydov/blob/f65f1848a17074280c5686eb7f106570bafe36fb/tests/test_util_owsutil.py#L21

I'll try to make something similar here, next to an online test.

cehbrecht:
Changes look reasonable to me. I will make a squash-merge.

@Roel 
The monkey-patch could be unnecessary when the `wfs` module be able to use stored XML documents like in the `wms` module:
https://github.com/geopython/OWSLib/blob/97cfddbfce6ea62655d7fdb2462d55770efde665/owslib/map/wms111.py#L77

cehbrecht:
We would like to have a ticket which can be closed by a PR, as described in the PyWPS project:
http://pywps.readthedocs.io/en/master/development.html

I have added on just for completeness: #431

cehbrecht:
@Roel 
merged :)

Roel:
Cool! Thanks!

PR #404, opened 2017-12-09
Add geometry column to schema

drnextgis:
Also add support of new MultiSurfacePropertyType type. Fiona compatibility is preserved.


drnextgis:
For building GetFeature requests with `propertyname` option we have to know name of geometry column. For now `get_schema` method doesn't return this important information. This PR fixes that flaw. @jachym any thoughts? As I can see you are the original author of the DescribeFeatureType parsing stuff.

PR #418, opened 2018-02-26
Support namespace to get 'crs' and 'dimension' attributes in Bounding…

Fifan31:
In a wps requests, if attributes 'crs' and 'dimension' are given with namespace in the input boundingBox description, these attributes are not taken into account.
With the proposed fix, 'crs' and 'dimension' attributes are read in both cases (with and without namespace).

### DIFF
--- a/owslib/ows.py
+++ b/owslib/ows.py
@@ -219,14 +219,14 @@ class BoundingBox(object):
         self.maxx = None
         self.maxy = None
 
-        val = elem.attrib.get('crs')
+        val = elem.attrib.get('crs') or elem.attrib.get('{%s}crs' % namespace)
         try:
             self.crs = crs.Crs(val)
         except (AttributeError, ValueError):
             LOGGER.warning('Invalid CRS %r. Expected integer')
             self.crs = None
 
-        val = elem.attrib.get('dimensions')
+        val = elem.attrib.get('dimensions') or elem.attrib.get('{%s}dimensions' % namespace)
         if val is not None:
             self.dimensions = int(util.testXMLValue(val, True))
         else:  # assume 2


owslib/ows.py
```diff
@@ -219,14 +219,14 @@ def __init__(self, elem, namespace=DEFAULT_OWS_NAMESPACE):
         self.maxx = None
         self.maxy = None
 
-        val = elem.attrib.get('crs')
+        val = elem.attrib.get('crs') or elem.attrib.get('{%s}crs' % namespace)
```
tomkralidis: Is it possible to update to `elem.attrib.get('{{}}crs'.format(namespace))` to be more Python 3 friendly?

owslib/ows.py
```diff
@@ -219,14 +219,14 @@ def __init__(self, elem, namespace=DEFAULT_OWS_NAMESPACE):
...
         try:
             self.crs = crs.Crs(val)
         except (AttributeError, ValueError):
             LOGGER.warning('Invalid CRS %r. Expected integer')
             self.crs = None
 
-        val = elem.attrib.get('dimensions')
+        val = elem.attrib.get('dimensions') or elem.attrib.get('{%s}dimensions' % namespace)
```
tomkralidis: Here as well.

cehbrecht:
I have added a bbox test for `ows` to work on this (#465). But I haven't found an example with a `BoundingBox` and a namespace for the `crs` and `dimensions` attributes.

See: http://docs.opengeospatial.org/is/14-065/14-065.html#101

cehbrecht:
@Fifan31 If this patch is still relevant than please update this PR or provide a new one including a test.
See this test example:
https://github.com/geopython/OWSLib/blob/56410931e3234e46a077fb15c436294154c30a0c/tests/test_ows.py#L6

I close this PR for now.

PR #422, opened 2018-03-20
first pass fix CSW tests

tomkralidis:


tomkralidis:
Thanks @cehbrecht.  Fixed/updated.

cehbrecht:
@tomkralidis thanks for the changes.

Issue #339, opened 2016-08-02
test failures ...


cehbrecht:
most of the time there are tests which fail due to not available online resources. This makes it hard to see if a patch causes a failure ... you need to check all the tests.

When using pytest one can mark for example long running tests as "slow" or those with external resources as "online":

http://docs.pytest.org/en/latest/mark.html

Tests which are currently broken and can not be fixed could be skipped:

http://docs.pytest.org/en/latest/skipping.html

Should we start using these markers?

cehbrecht:
shouldn't we start to fix the test suite?

https://travis-ci.org/geopython/OWSLib/builds

johanvdw:
I'm also in favor of switching to pytest, and willing to spend some time doing so.

Any objections?

tomkralidis:
Thanks for the info. Comments:

- we are already using pytest as our test suite
- what value does skipping tests provide over simply removing them?

cehbrecht:
if you don't need the test you can remove it. But sometimes a test is broken, we would like to keep it, but we can't fix it because it uses an external service which is down, or it is just not the module we feel responsible for.  By skipping it, we don't block the tests ... still have a reminder and the chance to fix it.

pytest is calling the doctests ... don't know how to add the skip decorator here.

kwilcox:
We have needed to run or mock the services used in tests for a long time.

How about we mark them as "remote", setup travis to run "-m remote" and "-m not remote" and allow failures on the "-m remote" tests?

cehbrecht:
... I like the "remote" tag ... does someone want to make an example?

johanvdw:
Here is a simple example:

https://github.com/johanvdw/OWSLib/commit/8beb3134e19fa59f9fa9740be407a8595dc53975

The results should look like:

```
johan@x1:~/git/OWSLib$ py.test
========================================== test session starts ==========================================
platform linux2 -- Python 2.7.13, pytest-3.0.6, py-1.5.2, pluggy-0.4.0 -- /usr/bin/python
cachedir: .cache
rootdir: /home/johan/git/OWSLib, inifile: tox.ini
plugins: hypothesis-3.33.0
collected 1 items 

tests/test_sos.py::TestSos::test_content PASSED

======================================= 1 passed in 0.31 seconds ========================================
johan@x1:~/git/OWSLib$ py.test -m "not remote"
========================================== test session starts ==========================================
[...]

PR #429, opened 2018-03-23
Fix #428: using utils.service_ok helper in pytest.skipif as criteria

### LINKED ISSUE #428: using "utils.service_ok" in tests for pytest.skipif criteria
Add a helper function `utils.service_ok()` to check if a service is available. This function can be used in the `pytest.skipif` as criteria. This function handles also a `ReadTimeout` exception.

cehbrecht:
This PR fixes #428.

It adds a `service_ok` helper function to be used by `pytest.skipif` which handles also a `ReadTimeout` exception.

The "normal" tests (see #339) are changed to use this function.

Other changes:
* renamed `_broken/doctest` to `_broken/doctests`.
* convert `wms_jpl_capabilities` from doctest to "normal" test.


tomkralidis:
Do we have use for it in `owslib.util`? If we don't I would suggest we keep in `tests.util` until needs change?

cehbrecht:
Ok, let's leave it in `tests.utils`.

PR #438, opened 2018-04-06
Fix #437: add headers, verify and cert options to wps

### LINKED ISSUE #437: Add headers, verify and cert option to wps module
Add the `requests` options `headers`, `verify` and `cert` to the `wps` initialization:

     from owslib.wps import WebProcessingService
     wps = WebProcessingService("http://localhost/wps", 
                headers={'my-header': 'test'}, 
                verify=True, cert='cert.pem')

The `verify` and `cert` option are used for SSL handling and passed to the `requests` library. See #435 and #340.

The `headers` option is also passed to the `requests` library and was introduced already in the `wms` module #321.  

See also #315 for a discussion how to pass the parameters to the `requests` library which is used in `util.openURL`.

cehbrecht:
This PR fixes #437.

Changes:
Add `headers`, `verify` and `cert` options to `wps` module which are passed to the `requests` library called by `util.openURL`.

### DIFF
--- a/owslib/wps.py
+++ b/owslib/wps.py
@@ -188,7 +188,8 @@ class WebProcessingService(object):
     Implements IWebProcessingService.
     """
 
-    def __init__(self, url, version=WPS_DEFAULT_VERSION, username=None, password=None, verbose=False, skip_caps=False):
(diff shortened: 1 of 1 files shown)


cehbrecht:
merged. 

We have discussed this issue on the coding-sprint in Bonn.

PR #449, opened 2018-04-24
Fix #333: add overwrite option to build_get_url

### LINKED ISSUE #333: usage of util.build_get_url()?
the params given to the `util.build_get_url()` function do not overwrite the query params from the `base_url`. Is this an intended behaviour?

For example when one inits a wps service with a caps url and builds a `describeprocess` url one gets the following:

```
from owslib.util import build_get_url
url = "http://localhost/wps?service=wps&request=getcapabilities"                                                                                                       
req_url = build_get_url(url, {'version':'1.0.0', 'request': 'describeprocess', 'identifier':'dummy'})
print req_url                                                                       
'http://localhost/wps?service=wps&request=getcapabilities&version=1.0.0&identifier=dummy'
```

The request parameter is not overwritten (still `getcapabilities` and not `describeprocess` as expected).

Should i add an optional `overwrite` flag to the function?

cehbrecht:
This PR fixes #333. 

Changes:
* adds on overwrite flag to `util.build_get_url` to allow overwrite of `base_url` parameters
* adds a test for `build_get_url`

### DIFF
--- a/owslib/util.py
+++ b/owslib/util.py
@@ -479,21 +479,42 @@ def getNamespace(element):
     else:
         return ""
 
-def build_get_url(base_url, params):
-    ''' Utility function to build a full HTTP GET URL from the service base URL and a dictionary of HTTP parameters. '''
 
-    qs = []
+def build_get_url(base_url, params, overwrite=False):
+    ''' Utility function to build a full HTTP GET URL from the service base URL and a dictionary of HTTP parameters.
+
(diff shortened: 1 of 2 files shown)


cehbrecht:
@tomkralidis This PR is compliant with the previous behavior of `util.build_get_url`. If you would like to have it in another way I can modify it with a new PR.

PR #450, opened 2018-04-24
Fix wps checkStatus etc

cehbrecht:
This PR has some combined minor changes to the wps module.

Changes:
* added common wps test
* using overwrite flag in `build_get_url`
* added getOperationByName method
* parse `creationTime` attribute in `checkStatus`
* handle xml parser exception in `checkStatus`


cehbrecht:
one doctest `sos_10_ndbc_getobservation.txt` timed out ... all others ok.

PR #483, opened 2018-05-28
Basic OWS Context implementation for ATOM XML and GeoJSON encodings

allixender:
Basic OWS Context implementation for ATOM XML and GeoJSON encodings.

It is possible to read/parse OWS Context documents in both ATOM XML and GeoJSON formats.

http://www.owscontext.org/

owc atom decode ATOM/XML encode and decode works except automatic geometry/bbox object representation, might not be needed so extensive support (would require extensive GEOJSON and GML parsing, which doesn't seem to be the immediate focus of OWSlib her)


tomLandry:
I wasn't able to provide you with any sort of feedback on this. We'll test in the summer and get back to you.

cehbrecht:
@tomkralidis I'm not to much familiar with OWSContext yet so I can't give a good review by myself. Shall we wait until a review is done before we accept this PR? Or shall we merge it (when it technically works) and mark it as alpha in the docs?

@allixender @tomkralidis Not sure how strict we want to be. Usually we would like to have PRs with one or just a few commits (I have learned that in pywps). But I could also do a "squash" merge in this case. 

@allixender Are all the example resources needed?

tomkralidis:
@cehbrecht thanks for the info.  This is a significant piece of work.  IMHO I took a quick and am +1 to merge and it appears to be a fulsome implementation.  +1 to mark alpha (or beta)?

I would also squash the merge into a single commit so we can revert if needed in a clean manner.

cehbrecht:
@allixender Would you like to update the docs with OWS Context and mark it something like alpha/beta/in-review? Some modules just reference a doctest for documentation. I can "squash" merge the PR then.

allixender:
@cehbrecht Where are the docs actually and how are they build? Happy to add examples to docs and add the beta/review text.

@cehbrecht  "Are all the example resources needed?", no actually not, I thought people might be interested as there are currently not many example files available in the web. I can of course reduce to a handful each, 3-5?

@tomkralidis Good hint with the "one commit", after I updated the docs I can create a new pull request with one commit/or Carsten does it?

cehbrecht:
@allixender The docs are just here:
https://github.com/geopython/OWSLib/blob/master/docs/en/index.rst

You can add the OWS Context to the standards list. Some of the other modules only reference a doctest for documentation. I leave this to you :)

I will do a "squash" merge on github of your pull request ... it will be only one commit in the master. So, you don't need to worry ... you can continue to commit to the current PR.

If the examples are created by you and/or not available elsewhere then better keep them. I thought we could reference some external examples in the docs.

allixender:
@cehbrecht Added the docs and other recommendations. I tried a rebase locally, but I am not sure if I did it right (doesn't look like it in the PR). I'd appreciat if you still could correct the squashing etc.
Looking forward.
Cheers,
Alex

cehbrecht:
done :) Thanks @allixender.  

I had to restart the test-suite several times to get finally a green light.

tomLandry:
Again, just checking if discussion from https://github.com/geopython/pywps/issues/322 should be ported here. We now have official and current OWS Context files from OGC testbed that could be used. Here at CRIM we can't help much in the next 4 to 6 weeks.

PR #485, opened 2018-06-04
added str and repr methods to ServiceIdentification and OperationsMetadata

huard:
Similar to what you've done, but for other classes.


cehbrecht:
@tomkralidis any objections from your side? Can I merge?

Issue #496, opened 2018-08-02
doctest sos_20_bom_gov_au.txt is a flickering test


cehbrecht:
... needs to be converted to a "normal" test with `skipif` check for the service availability.

cehbrecht:
Test was converted to "normal" test with PR #499. 

Still I had to skip the test due to a service exception:

```
Invalid temporalFilter, please specify a time instant or period for the field 'om:phenomenonTime'.
```

PR #497, opened 2018-08-02
Fix 494 wcs 200 tests

doclements:
A fix for failing test in the WCS 2 classes. The service used now is a reference implementation and as such should not change location etc.


cehbrecht:
Failures due to #496.

cehbrecht:
@doclements merged. Thanks :)

PR #498, opened 2018-08-02
Fix WPS offline tests

cehbrecht:
added `skip_caps=True` to avoid connecting to WPS service.


cehbrecht:
Failures due to #496.

Issue #491, opened 2018-07-23
ISO metadata parsing error when MD_FeatureCatalogueDescription uuidref attribute is missing


bchartier:
I use OWSLib in a tool checking metadata records stored in a CSW catalogue (https://github.com/geobretagne/mdchecker).
When it comes to parse ISO 19139 metadata records containing a <gmd:featureCatalogueCitation> element with no uuidref attribute an exception is raised by OWSLib (0.16.0):

```
Traceback (most recent call last):
  File "/Users/benji/dev/projects_current/mdchecker/mdchecker/venv/lib/python2.7/site-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/Users/benji/dev/projects_current/mdchecker/mdchecker/venv/lib/python2.7/site-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/Users/benji/dev/projects_current/mdchecker/mdchecker/venv/lib/python2.7/site-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/benji/dev/projects_current/mdchecker/mdchecker/venv/lib/python2.7/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/benji/dev/projects_current/mdchecker/mdchecker/venv/lib/python2.7/site-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/Users/benji/dev/projects_current/mdchecker/mdchecker/app/mdchecker/main.py", line 510, in quick_test
    metadatas, count, score = ins_wrapper.run_unrecorded_tests()
  File "/Users/benji/dev/projects_current/mdchecker/mdchecker/app/mdchecker/main.py", line 290, in run_unrecorded_tests
    maxharvest=self.test_params["maxharvest"]
  File "/Users/benji/dev/projects_current/mdchecker/mdchecker/app/mdchecker/inspirobot/Inspirobot.py", line 243, in mdsearch
    maxrecords=maxrecords, outputschema=self.OUTPUTSCHEMA)
  File "/Users/benji/dev/projects_current/mdchecker/mdchecker/venv/lib/python2.7/site-packages/owslib/csw.py", line 399, in getrecords2
    self._parserecords(outputschema, esn)
[...]

PR #502, opened 2018-08-14
Fix ISO metadata parsing for empty gmd:featureCatalogueCitation.

Roel:
This commit fixes the ISO metadata parsing for records that contain an
empty gmd:featureCatalogueCitation element.

It also adds tests for ISO metadata parsing as to ensure not to break
existing functionality in the future. Tests are added for two ISO
metadata records: one from DOV (already existing in the testdata) and
one from GéoBretage.

Closes issue #491.


cehbrecht:
@roel thanks for contributing this PR :) Maybe the `get_md_resource()` method in your test is a candidate for `tests/util.py`. But that can be changed later when we want to reuse this method in other tests.

Issue #493, opened 2018-07-27
XMLSyntaxError when format is JSON in CSW request


jankovicgd:
I am trying to get a record in JSON but getting a XMLSyntaxError. The default format works. Sending a simple GET with `outputFormat=application/json` also works.

### Input

```
csw = CatalogueServiceWeb(url)
csw.getrecordbyid(id=[id], 
                  outputschema='http://www.isotc211.org/2005/gmd', 
                  format='application/json')
```

### Output

```
Traceback (most recent call last):
  File "/home/nikola/.local/lib/python3.6/site-packages/IPython/core/interactiveshell.py", line 2963, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-3-ae9f74e2d058>", line 2, in <module>
    outputschema='http://www.isotc211.org/2005/gmd', format='application/json')
  File "/usr/lib/python3/dist-packages/owslib/csw.py", line 301, in getrecordbyid
    self._invoke()
  File "/usr/lib/python3/dist-packages/owslib/csw.py", line 680, in _invoke
    self._exml = etree.parse(BytesIO(self.response))
  File "src/lxml/etree.pyx", line 3426, in lxml.etree.parse
  File "src/lxml/parser.pxi", line 1856, in lxml.etree._parseDocument
  File "src/lxml/parser.pxi", line 1876, in lxml.etree._parseMemoryDocument
  File "src/lxml/parser.pxi", line 1764, in lxml.etree._parseDoc
  File "src/lxml/parser.pxi", line 1126, in lxml.etree._BaseParser._parseDoc
  File "src/lxml/parser.pxi", line 600, in lxml.etree._ParserContext._handleParseResultDoc
  File "src/lxml/parser.pxi", line 710, in lxml.etree._handleParseResult
  File "src/lxml/parser.pxi", line 639, in lxml.etree._raiseParseError
  File "<string>", line 1
XMLSyntaxError: Start tag expected, '<' not found, line 1, column 1
```

### Additional info

Running on Ubuntu 18.04 in a jupyter notebook with python 3.6.5.

Thanks in advance.

tomkralidis:
Is there a CSW we can test against?

jankovicgd:
Thanks for the fast reply. 

I am connecting at https://csw.eodc.eu/
Below is an ID. 
`M20170123_202730--_SIG0-----_S1BIWGRDH1VVD_104_A0101_OC010M_E070N065T1`

[...]

PR #511, opened 2018-08-19
[WIP] handle non-XML responses

tomkralidis:


cehbrecht:
This PR fixes #493.

cehbrecht:
@tomkralidis I have tried with the test mentioned in #493. Python 2.7 might work (didn't check) but Python 3.x does not. The response is in bytes (not string):
`response = b'{"csw:GetRecordByIdResponse": "..."}'`

So, we need some decode method handling python 2.x and python 3.x (`util.decode_sring()`). For Python 3.6 this converts bytes to string:
```
response.decode('utf-8')
'{"csw:GetRecordByIdResponse": "1"}'
```

OR

we may use a json loads test ... or relying on the mime type?
```
try:
    json.loads(response)
    is_json = True
except Exception
    is_json= False
```

We could add the mention test in #493 to the test suite.

cehbrecht:
@tomkralidis do you have an update on this PR? Though the status of owslib is not perfect do we want to got for a release 0.17?

tomkralidis:
@cehbrecht I haven't had a chance to follow up on this.  Pushing to 0.18.

PR #518, opened 2018-09-19
ship license

ocefpaf:
The license was renamed but the change was not updated here. The latest source distribution does not ship the license.


cehbrecht:
@ocefpaf thanks :)

Issue #521, opened 2018-10-20
add support for WFS 3


tomkralidis:
Add support for [WFS 3](https://github.com/opengeospatial/WFS_FES), which is a clean break from previous OGC service architecture (related: https://github.com/geopython/GeoHealthCheck/issues/221)

cehbrecht:
fixed by PR #524.

PR #524, opened 2018-11-04
add support for WFS 3 (#521)

tomkralidis:
This PR adds support for the [OGC Web Feature Service 3.0: Part 1 - Core](https://cdn.rawgit.com/opengeospatial/WFS_FES/3.0.0-draft.1/docs/17-069.html#_feature_collections) draft.

WFS3 is a clean break from traditional OGC service architecture (towards RESTful, JSON) and the code follows the same pattern (the only exception being the factory/constructor).

### DIFF
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -46,7 +46,7 @@ Standards Support
 +-------------------+---------------------+
 | `OGC CSW`_        | 2.0.2               |
 +-------------------+---------------------+
-| `OGC WPS`_        | 1.0.0               |
+| `OGC WPS`_        | 1.0.0, 1.1.0, 2.0, 3.0 |
 +-------------------+---------------------+
 | `OGC Filter`_     | 1.1.0               |
 +-------------------+---------------------+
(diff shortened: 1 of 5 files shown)


cehbrecht:
@tomkralidis I have reviewed your PR. Code looks simple and straight-forward. I followed successfully your examples. But I'm not a WFS user currently ... so there are probably better candidates as reviewer.

Some comments:
* Could you please check the codacy complains ... for example the `import` order for `six`.
* If you skip the `version` parameter the WFS init will fail. Could the version negotiation be handled in the WFS initialization?
* The `www.ldproxy.nrw.de` service does not return `numberMatched` ... well, not a client issue and I didn't check if this is optional.

cehbrecht:
@tomkralidis You could also update the docs for WFS-3. You could reference the tests as examples ... or we might still use `doctest` for documentation purposes.

tomkralidis:
Thanks @cehbrecht.  Updates pushed and rebased:

> Could you please check the codacy complains ... for example the import order for six.

Done

> If you skip the version parameter the WFS init will fail. Could the version negotiation be handled in the WFS initialization?

I would say this is by design (clean break)?

> The www.ldproxy.nrw.de service does not return numberMatched ... well, not a client issue and I didn't check if this is optional

The WFS 3.0 code does no parsing of the response (it's just JSON cast into a dict)

> You could also update the docs for WFS-3. You could reference the tests as examples ... or we might still use doctest for documentation purposes

Done.

cehbrecht:
test failures not caused by this PR.

cehbrecht:
@tomkralidis done :)

Issue #522, opened 2018-11-02
StoreSupported changes after describeprocess


huard:
Calling `describeprocess` changes the value of the storeSupported and statusSupported attributes of the process to True, even if the actual process definition is False. 

With version 0.17.0

```python
>>> from owslib.wps import WebProcessingService
>>> url = "http://localhost:5000/wps"  # Emu server
>>> wps = WebProcessingService(url)
>>> print(wps.processes[2])
WPS Process: nap, title=Afternoon Nap (supports sync calls only)
>>> print(wps.processes[2].storeSupported)
False
>>> wps.describeprocess('nap')
>>> print(wps.processes[2].storeSupported)
True
```

cehbrecht:
@huard Here is the code in owslib for setting `storeSupported`:
https://github.com/geopython/OWSLib/blob/609210002b676a15ccec8be3b80a8455cb313666/owslib/wps.py#L1450

It looks right but it is wrong ;) `storeSupported` is true when a non-empty string is returned (no matter "false" or "true") and false otherwise.

With a `GetCapabilities` request you don't get the `storeSupported` information, only with `DescribeProcess`. Need to check the specs if this is a bug in PyWPS.

huard:
I would expect the answer to be None if it is unknown (not included in the response), or trigger a call to describeProcess to get the actual boolean value.

cehbrecht:
fixed by PR #525.

PR #525, opened 2018-11-12
fix wps process properties (#522)

cehbrecht:
This PR fixes issue #522.

Fixes parser for WPS properties (`statusSupported`, `storeSupported`).


cehbrecht:
test failures not related to this PR.

PR #526, opened 2018-11-13
fix wps percentCompleted

cehbrecht:
This PR sets the WPS `percentCompleted` value to 100 when the process has succeeded.


cehbrecht:
See discussion in:
https://github.com/bird-house/emu/issues/62

cehbrecht:
@huard the wps service does not set a `percentCompleted` value when the process was successful ... it implies to be at 100 percent. This PR fixes the value accordingly in OWSLib.

PR #528, opened 2018-11-14
added bumpversion (#517)

### LINKED ISSUE #517: Use bumpversion
See:
https://pypi.org/project/bumpversion/

Simplifies the release process.

cehbrecht:
This PR adds `bumpversion` to make a new version of OWSLib. Docs are updated with an example how to use it.

Fixes #517.

### DIFF
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -569,6 +569,20 @@ Or ...
     # additional pep8 tests
     pep8 owslib/wmts.py
 
+Bump a new version
+------------------
+
+Make a new version of OWSLib in the following steps:
+
+* Make sure everything is commited to GitHub.
+* Update ``CHANGES.rst`` with the next version.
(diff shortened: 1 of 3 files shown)


cehbrecht:
@tomkralidis Are you happy with using bumpversion? May you could extend the docs with the release process so we can share it :)

tomkralidis:
@cehbrecht thanks for the info.  How does this make our release life easier, compared to f3aa5d2af33884d8fddf0dd14d112b8eda43fc84 for example?

cehbrecht:
@tomkralidis ping :)

See also our issue here (not related to `bumpversion` but where to place the version):
https://github.com/bird-house/emu/pull/67

We have decided to use the `requests` style:
https://github.com/requests/requests/blob/57d7284c1a245cf9fbcecb594f50471d86e879f7/setup.py#L61

cehbrecht:
... maybe for later ;)

PR #529, opened 2018-11-28
changed syntax to get rid of warning with python 3.6

huard:


cehbrecht:
@huard nice, thanks :)

PR #533, opened 2018-12-08
Fix GetFeature Request  URL with parameter MAP in MapServer and QGIS Server

samtux:
Better handling of URL parameters. Error for URL with MAP parameter in Mapserver and QGIS SERVER


samtux:
See issue https://github.com/geopython/OWSLib/issues/534

geographika:
@samtux - should be fixed with #612. Could you check this resolves your case?

geographika:
Closing as should be fixed in 0.32.0 release

Issue #536, opened 2018-12-12
WPS output as_reference depends on async


huard:
owslib.wps behaves in an unexpected and undocumented way. 

In principle, when composing an execute request, we can pass a list of output identifiers and whether or not we want the reference (True) or the data (False). 

When the execute request is create, this user preference is however overriden depending on the value of the mode (`_async`). So for example if I explicitly request a reference, but mode is SYNC, then the data will still be embedded in the response. 

My expectation would be for the default asReference value to depend on the server default, not OWSLib's default. 

```
            if isinstance(output, str):
                self._add_output(
                    responseDocumentElement, output, asReference=_async)
            elif isinstance(output, list):
                for (identifier, as_reference) in output:
                    self._add_output(
                        responseDocumentElement, identifier, asReference=as_reference and _async)
```

cehbrecht:
Should we fix it this way, that `asReference` is optional (which means the service decides):
```
# backward compatiblity for a single output
self._add_output(responseDocumentElement, output, asReference=None)
# list of output params with as_reference in True, False, None
self._add_output(responseDocumentElement, output, asReference=as_reference)
```

huard:
I like that. If asReference is True and storeSupported is False, then we can raise an error.

cehbrecht:
... it might be the issue, that we may not have the info `storeSupported` from the service. I will check.

huard:
My guess is that we may not have this from the getCapability, we'd have to call describeProcess to get it.

cehbrecht:
Yes, that is right:
```
<ProcessDescription wps:processVersion="1.5" storeSupported="true" statusSupported="true">
```

cehbrecht:
I can open a PR and make `as_reference` optional. Then we can still decide if owslib should run an extra request to figure out about this parameter.

cehbrecht:
@huard started a PR. Wherever you touch this module it doesn't feel right ... but we can't change this now.

I would not run a extra `DescribeProcess` request to get the `storeSupported` parameter. OWSLib is a low level library ... only requests demanded by the user (or the protocol) should be run. In case it does not fit, we get an exception message from the service. Birdy could handle this more nicely.

cehbrecht:
fixed by PR #537.

PR #537, opened 2018-12-13
make as_reference attribute optional (#536)

cehbrecht:
This PR fixes issue #536.

It makes the `as_reference` attribute of an output parameter optional.


cehbrecht:
test failure due to service timeout (wfs) ... not related to this PR.

PR #538, opened 2018-12-15
added correct checks, defaults and conversion to REST only WMTS

### LINKED ISSUE #461: Integer row/col failing with wmts.gettile
We have found in Cartopy that we need to convert row/col from integers to string representations for the WMTS gettile method. I don't know if this is the intent, or if integers should be expected to work, but I thought I would start an issue here.

We work around in Cartopy with the following change minor change: https://github.com/SciTools/cartopy/pull/1052

chrisgraf:
this also fixes #461


cehbrecht:
@tomkralidis I merge this one before #620 and update the PR.

cehbrecht:
@chrisgraf merged. It took a while. Sorry.

PR #542, opened 2018-12-18
is_reference returns False if value is binary type

### LINKED ISSUE #541: Raw data support in Py3
`owslib.wps.is_reference` returns True if the file is binary, because `b'' != ''`. This causes issues when trying to pass raw binary (b64) data in a ComplexDataInput, because the data is interpreted as a reference.

huard:
Adds a check for binary types in `is_reference`.

Fixes #541 

TODO: Add a test

### DIFF
--- a/owslib/wps.py
+++ b/owslib/wps.py
@@ -121,6 +121,7 @@ try:                    # Python 3
 except ImportError:     # Python 2
     from urlparse import urlparse
 
+import six
 # namespace definition
 n = Namespaces()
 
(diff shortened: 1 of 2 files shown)


cehbrecht:
Ok, I see. The issue was just the test:
```
parsed.scheme != ''
```
So, we are still relying on `urlparse` ... but the check must be py2/py3 safe.

cehbrecht:
I think we can test the `is_reference` in:
https://github.com/geopython/OWSLib/blob/master/tests/test_wps.py

... with examples we expect (py2, py3).

PR #540, opened 2018-12-17
Added support of setting output in ResponseDocument mimetype

eran-pinhas:
solving issue #539

Issue #539, opened 2018-12-16
Option to add mimetype to wps request


eran-pinhas:
Hey guys

https://github.com/geopython/OWSLib/blob/916003071191b9c0a94a8d9a352bd87e42ab0e43/owslib/wps.py#L708-L711

in these lines the WPS creates the xml element that we will send as WPS execute payload as the output details.

Will it be possible to declare the mime type of the output? I thought of something like this:

```python
 elif isinstance(output, list): 
     for ouputTuple in output: 
        if(len(tuple) == 2):
          (identifier, as_reference) = tuple
        else:
          (identifier, as_reference, mimeType) = tuple
         # of course adjusting _add_output to handle the mimeType will be required
         self._add_output( 
             responseDocumentElement, identifier, asReference=as_reference, mimeType=mimeType) 
```
*In this way the backward compatibility

any opinions?

eran-pinhas:
@cehbrecht

cehbrecht:
fixed by PR #540.

PR #546, opened 2019-01-01
Examples update: mimeType of output from recent api change

eran-pinhas:
As issued at #539 and fixed at #540


cehbrecht:
@eran-pinhas thanks :) I have forgotten the examples. Maybe we should use a doctest.

Issue #520, opened 2018-10-08
Is calling __init__ from __new__ in  OWSLib/owslib/feature/wfs110.py necessary?


vjf:
I have been trying to 'pickle' instances of "WebFeatureService_1_1_0", but this is difficult because the code calls \_\_init\_\_ from within \_\_new\_\_.

When unpickling the object, 'pickle' calls \_\_new\_\_ to create the instance

(From https://docs.python.org/3/library/pickle.html#pickling-class-instances)

_def load(cls, attributes):
    obj = cls.\_\_new\_\_(cls)
    obj.\_\_dict\_\_.update(attributes)
    return obj_
    
then because this calls \_\_init\_\_, it tries to fully re-create the object, which makes the use of pickle pointless!    
    
In the end, I had to override WebFeatureService_1_1_0.\_\_new\_\_ with one that does not call \_\_init\_\_ to get pickle to work.

Also I understand that python calls \_\_new\_\_ then calls \_\_init\_\_ when creating objects, (https://docs.python.org/3/reference/datamodel.html#object.__new__)

This means that \_\_init\_\_ is called twice!

Is this a bug, or is there a good reason for calling \_\_init\_\_ from within \_\_new\_\_? 

If it is a bug, I'd be happy to submit a pull request for this and the other WFS versions.

geographika:
Resolved with #548

geographika:
@vjf - do you have an example of pickling the WFS class? I get errors as lxml classes can't be pickled:

```python
import pickle
from owslib.wfs import WebFeatureService

wfs_url = "https://demo.mapserver.org/cgi-bin/wfs?"
wfs = WebFeatureService(url=wfs_url, version="2.0.0")

with open("wfs_object.pkl", "wb") as f:
    pickle.dump(wfs, f)
    # TypeError: cannot pickle 'lxml.etree._Element' object
```

Just wondering what approach you used (the `dill` library also fails).

vjf:
@geographika 

It was a long time ago, I think I was using 'diskcache'

This will fail with a TypeError when calling 'pickle.load()' if OWSLib < v0.32.1:

```
from owslib.wfs import WebFeatureService
from diskcache import Cache
import copyreg
from lxml import etree

wfs_url = "https://demo.mapserver.org/cgi-bin/wfs?"
wfs_dk = WebFeatureService(url=wfs_url, version="2.0.0")


def element_unpickler(data):
    return etree.fromstring(data)

def element_pickler(element):
[...]

PR #548, opened 2019-01-07
Strip out redundant __new__ to enable object pickling

vjf:
I'd like to remove the redundant '__ new __' from the wfs classes in order to fix two problems:
1. 'pickle' does not work (you cannot pickle the WFS classes)
2. '__ init __' gets called twice each time the class is initialised. This creates a substantial performance lag on slower network connections.

See https://github.com/geopython/OWSLib/issues/520 for more information.

Thanks.

### DIFF
--- a/owslib/feature/wfs100.py
+++ b/owslib/feature/wfs100.py
@@ -73,48 +73,6 @@ class WebFeatureService_1_0_0(object):
 
     Implements IWebFeatureService.
     """
-
-    def __new__(
-        self,
-        url,
-        version,
-        xml,
-        parse_remote_metadata=False,
-        timeout=30,
-        headers=None,
-        username=None,
-        password=None,
-        auth=None,
-    ):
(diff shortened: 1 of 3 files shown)


vjf:
Will try again.

geographika:
Reopening to check if this is still the case.

geographika:
The `__new__` overrides date back to at least 15 years - https://github.com/geopython/OWSLib/commit/e1d37a1df7b4afa5b558c87c5a84c270310c47e5
Since then OWSLib has moved from Python2 to Python3, I'm sure there was a reason at the time, but I can't see a need for the override now. Removing simplifies the code, removes duplicate calls to servers and resolves associated performance issues. Will be merging in the next 7 days - will leave open for comment until then. 

Thanks @vjf for the fix.

vjf:
@geographika @landryb Thanks.

PR #549, opened 2019-01-08
Add mimetype, schema and encoding to references

### LINKED ISSUE #547: WPS reference missing mimetype, encoding and schema
According to table 25 of the [WPS spec](http://docs.opengeospatial.org/is/14-065/14-065.html#45),  references should at least include the mimetype (defined in table 24) but it's currently missing in the implementation (https://github.com/geopython/OWSLib/blob/master/owslib/wps.py#L1605)

enolfc:
Fixes #547.


cehbrecht:
@enolfc thanks :)

Issue #550, opened 2019-01-08
tests on travis fail due to pytest version conflict


cehbrecht:
... here is part of an exception message:
```
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/pluggy/__init__.py", line 402, in load_setuptools_entrypoints
    "Plugin %r could not be loaded: %s!" % (ep.name, e))
pluggy.PluginValidationError: Plugin 'pytest_cov' could not be loaded: (pytest 3.3.0 (/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages), Requirement.parse('pytest>=3.6'))!
The command "python -m pytest" exited with 1.
```

cehbrecht:
fixed by PR #551.

PR #551, opened 2019-01-08
use pytest>=3.6

cehbrecht:
This PR fixes issue #550.

Added version restriction for pywps.


cehbrecht:
`tests/test_csw_ngdc.py::test_csw_ngdc FAILED`
... due to unavailable noaa service used by test.

PR #557, opened 2019-01-10
fixes #556

### LINKED ISSUE #556: WMS.getmap failure when response header lacks a content-type
An ncWMS2 server we're using does not seem to include a content-type attribute in the response header, which is assumed to be present by the getmap methods in wms130 and wms111. 

Is this a server-side bug ?

huard:
Change in WMS response handling to make sure it works when content-type attribute is not in header.


cehbrecht:
I can't answer the question in #556 but this PR looks like a safe approach.

@tomkralidis any objections?

tomkralidis:
Fair enough.

cehbrecht:
@huard done :)

PR #559, opened 2019-01-14
Support namespace et get 'crs' and 'diemension' attribute in BoundingBox

Fifan31:
I resubmit this PR with remarks taken into account (usage of format) and a unit test.

In a wps requests, if attributes 'crs' and 'dimension' are given with namespace in the input boundingBox description, these attributes are not taken into account.
With the proposed fix, 'crs' and 'dimension' attributes are read in both cases (with and without namespace).

### DIFF
--- a/owslib/ows.py
+++ b/owslib/ows.py
@@ -235,14 +235,14 @@ class BoundingBox(object):
         self.maxx = None
         self.maxy = None
 
-        val = elem.attrib.get('crs')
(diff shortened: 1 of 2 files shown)


cehbrecht:
test failures due to timeouts.

cehbrecht:
@Fifan31 done. Thanks :)

Fifan31:
Great !

PR #562, opened 2019-02-27
select properly the XML url in csw invoke

### LINKED ISSUE #561: CatalogueServiceWeb.getrecords2 selects bad url
getrecords2 selects endpoint that accepts SOAP instead of XML

sample endpoint http://mapy.geoportal.gov.pl/wss/service/CSW/guest/CSWStartup
testing script:
```.py
from owslib.csw import CatalogueServiceWeb
csw_addr='http://mapy.geoportal.gov.pl/wss/service/CSW/guest/CSWStartup'
csw = CatalogueServiceWeb(csw_addr)
csw.getrecords2()
```

sobczyk:
should fix #561

### DIFF
--- a/owslib/csw.py
+++ b/owslib/csw.py
@@ -640,6 +640,7 @@ class CatalogueServiceWeb(object):
                 else:
                     post_verbs = [x for x in op.methods if x.get('type').lower() == 'post']
                     if len(post_verbs) > 1:
+                        const_found = False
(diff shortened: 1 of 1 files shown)


sobczyk:
This is a quick fix that works, the break did not work as intended.
Proper way would be to terminate the outer loop as soon as an xml url was found,
but I think this works slightly slower with only minimal changes.

PR #567, opened 2019-03-13
dataType using value, then parse reference as needed

fmigneault:
Executes a better parsing of `DataType` using first the value if provided, and then `reference` parsing if provided to attempt finding the underlying data type.
Doesn't only use the `:` split to avoid errors from references specified as literal URL.

ref #383
fix for crim-ca/weaver#11

### DIFF
--- a/owslib/wps.py
+++ b/owslib/wps.py
@@ -1080,9 +1080,11 @@ class InputOutput(object):
                     self.dataType = sub_element.text
                     if not self.dataType:
                         reference = sub_element.get(nspath("reference", ns=subns))
+                        # backward search of first non-alpha character (:, #, /, etc.)
                         pos = len(reference) - 1
                         while pos >= 0 and reference[pos].isalpha():
                             pos -= 1
+                        # obtain substring after found non-alpha character position
                         self.dataType = reference[pos + 1:]
 
             for sub_element in literal_data_element:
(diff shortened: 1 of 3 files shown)


owslib/wps.py
```diff
@@ -1077,9 +1077,13 @@ def _parseLiteralData(self, element, literalElementName):
             for sub_element in literal_data_element:
                 subns = getNamespace(sub_element)
                 if sub_element.tag.endswith('DataType'):
-                    reference = sub_element.get(nspath("reference", ns=subns)) or sub_element.text
-                    if reference and ':' in reference:
-                        self.dataType = reference.split(':')[-1]
+                    self.dataType = sub_element.text
+                    if not self.dataType:
+                        reference = sub_element.get(nspath("reference", ns=subns))
+                        pos = len(reference) - 1
```
cehbrecht: Could you add a comment like "backward search for non-alpha char # or :"

owslib/wps.py
```diff
@@ -1077,9 +1077,13 @@ def _parseLiteralData(self, element, literalElementName):
...
+                        while pos >= 0 and reference[pos].isalpha():
+                            pos -= 1
+                        self.dataType = reference[pos + 1:]
```
cehbrecht: Add comment "get substring after # or : for type: string, etc"

fmigneault: done in 02a3b66

owslib/wps.py
fmigneault: done in 02a3b66

cehbrecht:
@fmigneault merged. Thanks :)

PR #570, opened 2019-03-26
Remove Python 3.4 From Travis CI

Coop56:
Python 3.4 has reached it's end of life, remove it from the build matrix so that the Travis CI tests can pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: python
 python:
   - "2.7"
+  - "3.5"
   - "3.6"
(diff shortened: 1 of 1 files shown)


sbrunner:
Shouldn't we replace it by Python 3.5?

tomkralidis:
Can we test with adding 3.5 and 3.7?

Coop56:
Sure thing, I've added Python 3.5 as well.  3.7 was already added by a previous PR I sent in a few weeks ago.

PR #573, opened 2019-04-14
added the option to send timeout parameters to the readers in wps class

omerlevko:


cehbrecht:
@omerlevko merged. thanks :)

PR #575, opened 2019-05-13
Ensure authentication parameters are provided to WFS 3 requests.

bosth:
`auth` and `password` were set but not used with WFS 3 in OWSLib.


tomkralidis:
Thanks @bosth !

Issue #412, opened 2018-01-10
WCS WebCoverageService authentication support?


afrigeri:
Hello,  is there some way to include authentication  in the WebCoverageService class, like username and password parameters as  in WebMapService_1_1_1?

Thanks 

Alessandro

amirmohammadmir:
Hi. Did you find the answer?
Thanks

tomkralidis:
You can now add an [Authentication ](https://github.com/geopython/OWSLib/blob/master/owslib/util.py#L832)object as follows:

```python
from owslib.wcs import WebCoverageService
w = WebCoverageService('https://example.org/wcs', Authentication(username='foo', password='bar')
```

amirmohammadmir:
@tomkralidis  Unfortunately the [website](https://maps.isric.org/) I'm trying to connect is not working . I check your solution after that.

amirmohammadmir:
Here is the completed code (thanks to @tomkralidis): 
```
from owslib.util import Authentication
from owslib.wcs import WebCoverageService
wcs = WebCoverageService('http://maps.isric.org/mapserv?map=/map/phh2o.map', auth=Authentication(username='foo', password='bar'),
                         version='1.0.0')

bbox = (-1784000, 1356000, -1140000, 1863000)
response = wcs.getCoverage(
    identifier='phh2o_0-5cm_mean',
    crs='urn:ogc:def:crs:EPSG::152160',
    bbox=bbox,
    resx=250, resy=250,
    format='GEOTIFF_16')
with open('./data/Senegal_pH_0-5_mean.tif', 'wb') as file:
    file.write(response.read())

import rasterio
CEC = rasterio.open("./data/Senegal_pH_0-5_mean.tif", driver="GTiff")

from rasterio import plot
plot.show(CEC, title='Mean pH between 0 and 5 cm deep in Senegal', cmap='gist_ncar')
```

1 problem:

1. It gives me this error (do to the server errors):
`requests.exceptions.ReadTimeout: HTTPConnectionPool(host='maps.isric.org', port=80): Read timed out. (read timeout=30)
`
For now, I cannot create a username and password and check if this works or not.

amirmohammadmir:
The servers work just fine for now. I have a new problem and would appreciate any help from you,

[...]

Issue #489, opened 2018-06-28
OWSLib throws SSLError when retrieving GetCapabilities on valid WMS


erget:
I'm trying to build a hello world with the EUMETSAT EUMETView WMS. I know that the web service works - it's demonstrated [here](https://eumetview.eumetsat.int/mapviewer/) and has a human-readable description [here](https://www.eumetsat.int/website/home/Images/EUMETView/index.html). However, when I try to access it using OWSLib it throws an SSLError. Accessing the same document from my browser downloads the document without problems.

Here's the offending code:

```
from owslib.wms import WebMapService
wms = WebMapService("http://eumetview.eumetsat.int/geoserv/meteosat/wms/?service=WMS&request=GetCapabilities&version=1.1.1")
```

Which produces the following traceback:
```
---------------------------------------------------------------------------
Error                                     Traceback (most recent call last)
~/.miniconda3/lib/python3.6/site-packages/urllib3/contrib/pyopenssl.py in wrap_socket(self, sock, server_side, do_handshake_on_connect, suppress_ragged_eofs, server_hostname)
    440             try:
--> 441                 cnx.do_handshake()
    442             except OpenSSL.SSL.WantReadError:

~/.miniconda3/lib/python3.6/site-packages/OpenSSL/SSL.py in do_handshake(self)
   1906         result = _lib.SSL_do_handshake(self._ssl)
-> 1907         self._raise_ssl_error(self._ssl, result)
   1908 

~/.miniconda3/lib/python3.6/site-packages/OpenSSL/SSL.py in _raise_ssl_error(self, ssl, result)
   1638         else:
-> 1639             _raise_current_error()
   1640 

~/.miniconda3/lib/python3.6/site-packages/OpenSSL/_util.py in exception_from_error_queue(exception_type)
     53 
---> 54     raise exception_type(errors)
     55 

Error: [('SSL routines', 'ssl3_get_server_certificate', 'certificate verify failed')]

During handling of the above exception, another exception occurred:

SSLError                                  Traceback (most recent call last)
~/.miniconda3/lib/python3.6/site-packages/urllib3/connectionpool.py in urlopen(self, method, url, body, headers, retries, redirect, assert_same_host, timeout, pool_timeout, release_conn, chunked, body_pos, **response_kw)
    600                                                   body=body, headers=headers,
--> 601                                                   chunked=chunked)
    602 

[...]

PR #580, opened 2019-06-05
Improve Authentication Options

eric-spitler:
Add or improve upon authentication methods for WMS/WCS/WFS/WMTS/TMS services

In some cases, username/password authentication was partially there - service-level attributes set but capabilities-level not set. In other cases, no authentication was capable.

Changes:
- For `*Service`, `*Capabilities`, `ContentMetadata` classes, allow `username` | `password` | `cert` | `verify` arguments to `__init__()`
- For uses of `openURL`, include additional keyword arguments for `username` | `password` | `cert` | `verify`
- For uses of `requests` library, include additional keyword arguments for `username` | `password` | `cert` | `verify` (uses `auth=(username, password)`)
- When creating other class instances, include current values for these arguments

Default behavior for these arguments in the `requests` and `openURL` implementation is to default to `None`, which for username/password authentication means "unauthenticated".

For certificate-based authentication, defaulting to `cert=None` means unauthenticated. In the case of 2-way certificate authentication, paths to a cert/key combination is used as the `cert=` argument and the path a CA bundle is used for the `verify=` argument. If `verify=True` (default), use system CA bundle.


eric-spitler:
Should resolve https://github.com/geopython/OWSLib/issues/412

eric-spitler:
Enables use of `cert` and `verify` for https://github.com/geopython/OWSLib/issues/489

tomkralidis:
Thanks @eric-spitler .  I've asked @cehbrecht to review as well.  Can the changes incorporate CSW as well?  I'm also wondering whether we need a custom `owslib.util.auth` object which gets passed around instead of tacking on parameters across functions all the time.  Thoughts?

VERSION.txt
```diff
@@ -1 +1 @@
-0.17.1
+0.17.2
```
cehbrecht: Change version with next release?

cehbrecht:
Tests failures are not related to the PR.

cehbrecht:
> I'm also wondering whether we need a custom owslib.util.auth object which gets passed around instead of tacking on parameters across functions all the time. Thoughts?

@eric-spitler @tomkralidis Would be nice to simplify this code :) But that is probably for another PR. Could we use a base OWS class with an internal `openURL` method which has all the auth parameters from the class instance. Maybe that would avoid some of the parameter passing.

VERSION.txt
eric-spitler: Sure, I'll get that changed back in both files.

eric-spitler:
> > I'm also wondering whether we need a custom owslib.util.auth object which gets passed around instead of tacking on parameters across functions all the time. Thoughts?
> 
> @eric-spitler @tomkralidis Would be nice to simplify this code :) But that is probably for another PR. Could we use a base OWS class with an internal `openURL` method which has all the auth parameters from the class instance. Maybe that would avoid some of the parameter passing.

Completely agree. There so many places where the same changes were made. I was mostly trying to retain existing structure/functionality so as not to impact existing testing and utilization, but a consolidated solution would be best.

Also, some of the changes here are things I've done on a private project with local modifications, specifically around WMS/WCS and cert/verify usage. In that case I have a real site I'm accessing that requires that form of authentication, but I'd have to think about how to properly test that w/o relying on someone else's site/data.

eric-spitler:
Just a quick update, I'm working on a centralized authentication solution but have a couple test kinks to work out. Basically doing what you said with an `Authentication` class in `owslib.util` and updating existing uses of `username` and `password` to merge into the new class. That class then has `.openURL(*args, **kwargs)` and `.get(*args, **kwargs)` to handle authenticated requests.

tomkralidis:
Thanks @eric-spitler.  Should we wait for that work to come through in lieu of this PR?

eric-spitler:
Work got pretty busy lately so I don't know when I'll be able to finish the updates. I would say merge, but they'll make backwards compatibility messy. If you're okay waiting I'll keep going on it.

tomkralidis:
Perhaps it’s better to wait rather than introduce then replace functionality.

eric-spitler:
@tomkralidis @cehbrecht Updates are ready for your review.

owslib/csw.py
```diff
@@ -658,7 +660,7 @@ def _invoke(self):
 
         if isinstance(self.request, six.string_types):  # GET KVP
             self.request = '%s%s' % (bind_url(request_url), self.request)
-            self.response = openURL(self.request, None, 'Get', username=self.username, password=self.password, timeout=self.timeout).read()
+            self.response = self.auth.openURL(self.request, None, 'Get', timeout=self.timeout).read()
```
tomkralidis: Should `openURL` be a function of the new Authentication mechanism or should we do it the other way around (i.e. pass an auth object as a parameter to `openURL`)

owslib/csw.py
```diff
@@ -35,7 +33,7 @@
 from owslib.dif import DIF
 from owslib.gm03 import GM03
 from owslib.namespaces import Namespaces
-from owslib.util import cleanup_namespaces, bind_url, add_namespaces, openURL
+from owslib.util import cleanup_namespaces, bind_url, add_namespaces, OrderedDict, Authentication
```
tomkralidis: Why we adding `OrderedDict` here?

owslib/util.py
```diff
@@ -17,6 +18,7 @@
 from owslib.etree import etree, ParseError
 from owslib.namespaces import Namespaces
 from six.moves.urllib.parse import urlsplit, urlencode, urlparse, parse_qs, urlunparse
+import traceback
```
tomkralidis: Why are we adding `traceback` here?

tomkralidis:
thanks @eric-spitler for the impressive work here!  See my review/comments. In addition:

- can you fix merge conflicts?
- aside: I imagine given the nature of the enhancement it is difficut to provide tests against services requiring authentication.  It would be nice to have test services at some point to help test. This shouldn't stop this PR from being merged however once it is approved.

owslib/csw.py
eric-spitler: Not sure how that got added. I'll remove it.

owslib/util.py
eric-spitler: Not sure how that got added. I'll remove it.

owslib/csw.py
eric-spitler: Its up to you. I was thinking make the auth passing transparent, but can do it either way.

tomkralidis: @cehbrecht thoughts?

eric-spitler: Implemented

eric-spitler:
The tests had 500/502 responses from external locations. Is there a way to trigger a re-run or just a new commit?

Issue #578, opened 2019-06-05
WFS3 client: add api() function


justb4:
As per [WFS3 spec](http://docs.opengeospatial.org/DRAFTS/17-069r1.html):

![image](https://user-images.githubusercontent.com/582630/58948760-26a40c80-878b-11e9-867e-b81eb11f3895.png)

in `OWSLib 0.17.1` found function missing.

`wfs300.py` should add function like:

```
    def api(self):
        """
        implements Requirement 3 (/req/core/api-definition-op)

        @returns: api object
        """

        url = self._build_url('api')
        LOGGER.debug('Request: {}'.format(url))
        response = requests.get(url, headers=REQUEST_HEADERS).json()
        return response
```

justb4:
PR upcoming

PR #583, opened 2019-06-17
status set to 'Exception' inside _parseExceptionReport

DrMenny:
There's a bug in the monitorExecution() function:

if the execution.status is not _'None'_ and there's an ExceptionReport, the _monitorExecution()_ will be trapped inside the while loop, because the _parseExceptionReport() cannot set the execution.status to 'Exception'.

In the image you can see that the error is printed every 3 seconds and it gets bigger and bigger because the _parseExceptionReport() appends the error each time the function is called.

image

If the self.status is set as 'Exception' when the _parseExceptionReport() is called, the error is printed just once and the monitorExecution() is stopped.

### DIFF
--- a/owslib/wps.py
+++ b/owslib/wps.py
@@ -1844,6 +1844,7 @@ def monitorExecution(execution, sleepSecs=3, download=False, filepath=None):
     '''
     Convenience method to monitor the status of a WPS execution till it completes (succesfully or not),
     and write the output to file after a succesfull job completion.
+
     :param execution: WPSExecution instance
     :param int sleepSecs: number of seconds to sleep in between check status invocations
     :param download: True to download the output when the process terminates, False otherwise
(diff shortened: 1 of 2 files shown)


DrMenny:
I don't understand what's wrong with the build.

DrMenny:
I just tried to build with the original code (the actual geopython/OWSLib) but the build fails anyway.

`tests/test_wfs3_pygeoapi.py::test_wfs3_pygeoapi FAILED`

Can you please consider my pull request anyway? Thanks (I will push my changes now)

DrMenny:
Here's my last commit. I tested the code and it works. 

image

Thank you for your attention!

DrMenny:
> geopython/owslib#578 add WFS3 api() call with tests and doc #579 
implemented

cehbrecht:
@DrMenny merged. Thanks :)

DrMenny:
> @DrMenny merged. Thanks :)

Thank you :) Are you going to release v0.18 soon?

PR #589, opened 2019-07-29
Add WFS get_schema to docs

geographika:
As discovered via https://gis.stackexchange.com/questions/109269/how-to-call-the-method-of-describefeaturetype-in-owslib


cehbrecht:
@geographika merged. Thanks :)

PR #592, opened 2019-08-13
copy and paste ready log example

tilmanb:
I think its easier for users when they can just copy/paste the logging code, so here is a minimum working example to log to the console


cehbrecht:
@tilmanb thanks :)

Issue #591, opened 2019-08-12
WPS execution error


dpostolovski:
`execution.getOutput() ` fails on Python 3.7 and 3.6 with error: 

```
File "/Users/xxx/PycharmProjects/xxx/venv/lib/python3.6/site-packages/owslib/wps.py", line 857, in getOutput
    content = content + output_content
TypeError: can't concat str to bytes
```

cehbrecht:
@dpostolovski Thanks for reporting. I have identified some issues with the `bytes` encoding and made a PR #593. I'm not sure if this solves the complete issue. Could you please provide an example?

For testing I was using the Emu WPS:
https://github.com/bird-house/emu

... the `hello`, `chomsky` and `inout` processes.

dpostolovski:
`http://www.soda-is.com/service/wps` with process `get_cams_radiation`

cehbrecht:
@dpostolovski the wps service is maybe blocked by the firewall ... I can't reach it. Did you try with the patch?
https://github.com/geopython/OWSLib/pull/593/files

... or is the source code of the wps available?

dpostolovski:
The server seems to be currently down, so I can't test either. Will let you know when it comes up again.

dpostolovski:
@cehbrecht The service is up again. I confirm that the patch works in my case.

cehbrecht:
@dpostolovski thanks for testing. PR #593 is merged.

PR #593, opened 2019-08-13
fix getOutput

cehbrecht:
This PR fixes issue #591. It solves the `bytes` encoding issue.


cehbrecht:
Test failure not related to this PR.

PR #596, opened 2019-08-20
Update Service Provider for cases where ProviderSite-tag is provided …

jannefleischer:
…but doesn't contain any href link


cehbrecht:
@jannefleischer thanks :)

PR #598, opened 2019-08-23
Avoid trying to parse None CRS bounding box values

### LINKED ISSUE #403: Unexpected warnings about CRS
When I create a new `WebFeatureService` I get a lot of warnings like these:

```
Invalid CRS %r. Expected integer
Invalid CRS %r. Expected integer
Invalid CRS %r. Expected integer
Invalid CRS %r. Expected integer
...
```

It looks like we have at least three issues here

* String interpolation is not complete, see [here](https://github.com/geopython/OWSLib/blob/master/owslib/ows.py#L226)
* Non-integer CRS also should be valid
* `WGS84BoundingBox` does [not contains](https://github.com/geopython/OWSLib/blob/master/owslib/feature/wfs110.py#L300) information about CRS. It means that we will be get this warnings all the time

walkermatt:
When no crs attribute is present skip attempting to create a CRS
instance. No crs attribute is present for a WGS84 bounding box for
instance.

Resolves #403

### DIFF
--- a/owslib/ows.py
+++ b/owslib/ows.py
@@ -241,10 +241,12 @@ class BoundingBox(object):
         self.maxy = None
 
         val = elem.attrib.get('crs') or elem.attrib.get('{{{}}}crs'.format(namespace))
-        try:
-            self.crs = crs.Crs(val)
-        except (AttributeError, ValueError):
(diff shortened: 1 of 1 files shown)


cehbrecht:
@walkermatt merged. Thanks :)

PR #599, opened 2019-08-23
fix wps.describeprocess

cehbrecht:
This PR fixes `wps.describeprocess`. It filters the requested process from the available process list (before it returned just the first one). This is needed when an xml response document is given with several process descriptions in test cases.


cehbrecht:
@huard Maybe you like to have a look at it?

cehbrecht:
@huard thanks :)

huard:
Nice that you got the tests fixed. This was annoying.

PR #602, opened 2019-08-30
Add Support WMS 1.3 ScaleDenominator

sbrunner:


cehbrecht:
@tomkralidis ok to merge?

tomkralidis:
@sbrunner is it possible to explain the calculations?  Should OWSLib be performing these calculations should we leave to the downstream client?  I don't mind keeping them so long as:

1. they are documented
2. the values that are parsed from Capabilities XML are stored as properties

sbrunner:
@tomkralidis

* I do the calculation to make the result equivalent as with WMS 1.1.1

```python
INCH_TO_M = 0.0254  # standard inch to meter conversion
OGC_PIXEL_SIZE = 0.00028  # The OGC standard pixel size is 0.28 mm, here it's in meter
OGC_DPI = INCH_TO_M / OGC_PIXEL_SIZE  # Calculate the DPI
SCALEDENOM_TO_RESOLUTION = 1 / OGC_DPI * INCH_TO_M * sqrt(2)  # Calculate the resolution, the sqrt(2) is because the scale is on the diagonal and the resolution on the side
```

I don't know how the compatibility between 1.1.1 and 1.3.0 is important for you if it's not important I can simply remove the calculation...

CU
Stéphane

tomkralidis:
Thanks @sbrunner.  Can we keep `min_scale_denominator` and `max_scale_denominator` in the object (`self`) then?

sbrunner:
Done :-)

sbrunner:
Thanks @tomkralidis :-)

PR #611, opened 2019-09-26
fix auth in wps

### LINKED ISSUE #609: openURL not respecting Authentication object's verify property
When an Authorization object is passed to `openURL(...)` through the creation of a "service" object (WebFeatureService, WebMapService, etc.), the default value of `verify` overrides the value of `auth.verify`.  It is also notable that the Authorization object is mutable; Thus, `auth.verfiy` can be overwritten to True for subsequent use.

See the offending logic here: https://github.com/geopython/OWSLib/blob/03a0083e73690a43cd9a9c86fcf9107bc126f406/owslib/util.py#L174-L175

| verify  | auth.verify | verify & !auth.verify | ...result |
| :---: | :---: | :---: | :---: |
| T | T | F | `auth.verify` stays True |
| T | F | T | `auth.verify` gets overwritten to True |
| F | T | F | `auth.verify` stays True | 
| F | F | F | `auth.verify` stays False |

### Solutions:
1.  Only take Authorization objects in `openURL(...)`
Since you already have this object that contains all authorization values, let's use it!  Remove `username`, `password`, `verify`, and `cert` from the function. Every calling function would either need to create it's own Auth or pass one that was given to it. Since creating an Auth sets `verify` to True this defaults to being secure unless a function overwrites it. 

2. Remove `auth` from `openURL(...)`
Seems like a step backwards, but you could just make every calling function split out the contents of an Authorization object.

3. Change logic to respect any turning off of SSL verification. 
Logic: ``` if not verify or not auth.verify: verify = False
```.  Also, removes the issue of overwriting `auth.verify` to False when it's initially True (other properties will need similar treatment).

### PR DESCRIPTION
**Pull request by cehbrecht**
This PR fixes #609 in the wps module. It keeps the old interface with `verify` only for the `WebProcessingService` class. Internally only the `self.auth` parameter should be used.

I have made a quick test with [birdy](https://github.com/bird-house/birdy).

### DIFF
--- a/owslib/wps.py
+++ b/owslib/wps.py
@@ -115,6 +115,7 @@ from owslib.util import (testXMLValue, testXMLAttribute, build_get_url, clean_ow
 from xml.dom.minidom import parseString
 from owslib.namespaces import Namespaces
 from urllib.parse import urlparse
+import warnings
 
 # namespace definition
 n = Namespaces()
@@ -182,6 +183,24 @@ def is_complexdata(val):
     return isinstance(val, IComplexDataInput)
 
 
+def _fix_auth(auth, username=None, password=None, verify=None, cert=None):
+    """Updates auth from deprecated parameters username, password, verify and cert."""
+    if any(p is not None for p in (username, password, verify, cert)):
+        message = 'The use of "username", "password", "verify", and "cert" is deprecated. ' + \
+                  'Please use the "auth" keyword during class instantiation. ' + \
+                  'These keywords will be removed in a future release.'
+        warnings.warn(message, DeprecationWarning)
+    if username is not None:
+        auth.username = username
+    if password is not None:
+        auth.password = password
+    if verify is not None:
+        auth.verify = verify
+    if cert is not None:
+        auth.cert = cert
+    return auth
+
+
 class IComplexDataInput(object):
 
     """
@@ -213,7 +232,8 @@ class WebProcessingService(object):
 
         Parameters username, password, verify and cert are deprecated. Please use auth parameter.
         """
-        self.auth = auth or Authentication(username, password, cert, verify)
+        self.auth = auth or Authentication()
+        _fix_auth(self.auth, username, password, verify, cert)
 
         # fields passed in from object initializer
         self.url = clean_ows_url(url)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by cehbrecht** [comment]
@eric-spitler @tiranno Please give a comment.

**Comment by eric-spitler** [comment]
The WPS* classes already had the `username`|`password`|`cert`|`verify` arguments at the time of addition of the `Authentication` class. They were left in there for backward compatibility with the attempt to allow them to override the values in an `Authentication` instance (if needed) when using `readFromUrl` and such. It is certainly more consistent to enforce the use of `auth=`, but the changes here will be breaking changes for anyone using the original authentication methods.

There is a comment in the class docs about deprecation, but deprecated functionality needs to remain functional and produce warnings to the developer so they can make the necessary updates prior to removing the functionality altogether.

@cehbrecht - Does `owslib` have a development timeline for how long to keep deprecated functionality? Number of releases?

File: owslib/wps.py [hunk 1]
```diff
@@ -505,21 +489,17 @@ def __init__(self, version=WPS_DEFAULT_VERSION, verbose=False, timeout=None, aut
         super(WPSCapabilitiesReader, self).__init__(
             version=version, verbose=verbose, timeout=timeout, auth=auth)
 
-    def readFromUrl(self, url, username=None, password=None,
-                    headers=None, verify=True, cert=None):
+    def readFromUrl(self, url, headers=None):
```
eric-spitler: Should really issue a `DeprecationWarning` here for some time before removing existing functionality to avoid breaking code

```python
def readFromUrl(self, url, username=None, password=None, headers=None, verify=True, cert=None):
    import warnings
    message = 'The use of "username", "password", "verify", and "cert" is deprecated. ' + \
        'Please use the "auth" keyword during class instantiation. ' + \
        'These keywords will be removed in a future release.'
    warnings.warn(message, DeprecationWarning)
    # rest of the method
```

cehbrecht: I can add the deprecation warning in this PR. 

I would have considered the `WPSReader` class as part of the internal interface. Though this is not obvious from the code.

cehbrecht:
@tomkralidis Do you have recommendations on how to handle interface deprecations?

tomkralidis:
@cehbrecht we have had such warnings in other parts of OWSLib (example: https://github.com/geopython/OWSLib/commit/d4e26ea53e197b1a497aa8bf34a6ebf011158a96) for quite some time.  Technically OWSLib is a pre 1.0 project so we have not gone the route of stable branch support (everything happens straight and away in master).

All to say I would leave deprecation handling (how long, etc.) up to the maintainer of the given code.  We don't have an overall policy for OWSLib (yet) but perhaps this is a good time to make one).

cehbrecht:
@eric-spitler @tiranno @tomkralidis I have added the method `_fix_auth` which handles the deprecation warning. The interface is converted to the master branch version ... except that I use `verify=None` instead of `verify=True`.

cehbrecht:
@tomkralidis Can I merge? It fixes at least the WPS for HTTPS ...

tomkralidis:
@cehbrecht +1

Issue #534, opened 2018-12-08
Error GetFeature with parameter MAP in MapServer and QGIS Server


samtux:
For URLs that include map  parameter, such as path of project Mapserver or QGIS Server, the GetFeature (GET) operation constructs an erroneous request URL. Example:

```python
from owslib.wfs import WebFeatureService

url = "http://visorgeo.ambientebogota.gov.co/map/?map=/geodata/mapserver/mapfiles/calidad_aire/estacion.map"

wfs2 = WebFeatureService(url=url, version="2.0.0") 
wfs2.getGETGetFeatureRequest(typename="ms:estacion")

```
Returns a URL with parameter concatenation error (&?):

http://visorgeo.ambientebogota.gov.co/map/?map=/geodata/mapserver/mapfiles/calidad_aire/estacion.map&?service=WFS&version=2.0.0&request=GetFeature&typenames=ms:estacion

and the request:

```python
wfs2.getfeature(typename="ms:estacion").read()
```
with error:

    '<HTML>\n<HEAD><TITLE>MapServer Message</TITLE></HEAD>\n\n<BODY BGCOLOR="#FFFFFF">\nmapserv(): Web application error. Traditional BROWSE mode requires a TEMPLATE in the WEB section, but none was provided.\n</BODY></HTML>'

The suggested way to improve the use of parameters is to use the Python urlparse library.

https://github.com/geopython/OWSLib/pull/533

Issue #587, opened 2019-07-02
Enable providing and forwarding HTTP headers for endpoints


justb4:
Some OWS endpoint classes allow HTTP Headers via a `headers` dict structure to be passed on creation. For example [WebMapService](https://github.com/geopython/OWSLib/blob/master/owslib/wms.py#L23). But in this case these `headers` are only passed/forwarded when creating a WMS 1.3.0 instance.

For other OWS types like WFS, WMTS and TMS (and probably more) no `headers` can be passed. 

This is needed in cases where a client needs to pass specific  HTTP Header name/values, for example for Authentication via Bearer Token.

justb4:
I can work on this PR upcoming.

cehbrecht:
fixed by PR #613.

PR #612, opened 2019-10-09
WFS 2.0 URL building

geographika:
In WFS 2.0 GetFeature requests currently fail for MapServer URLs containing `?map=path/my.map&`
This is the same issue as #533 - and unfortunately I missed that there was also an open pull request at #534 from @samtux
The approach in this pull request is slightly different - it reuses the `build_get_url` function from utils. 

A few other notes:

- There is a warning about `owslib\util.py:561: DeprecationWarning: cgi.parse_qsl is deprecated, use urllib.parse.parse_qsl instead` so this has been updated
- the WFS GetFeature URL building allowed for the [doseq](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode)parameter to be used (I'm not sure when this applies to WFS requests, but I added it to avoid any breaking changes). It allows for automatic duplicate keys for sequences e.g.

```python
>>> urllib.urlencode({'keys': [1,2]}, doseq=True)
>>> 'keys=1&keys=2'
>>> urllib.urlencode({'keys': [1,2]}, doseq=False)
>>> 'keys=%5B1%2C+2%5D'
```
- I 've also added tests for the `doseq` and URLs with ampersands

### DIFF
--- a/owslib/feature/__init__.py
+++ b/owslib/feature/__init__.py
@@ -9,6 +9,7 @@ import logging
 
 from urllib.parse import urlencode
 from owslib.crs import Crs
+from owslib.util import log, Authentication, build_get_url
 from owslib.util import Authentication
 from owslib.feature.schema import get_schema
 from owslib.feature.postrequest import PostRequest_1_1_0, PostRequest_2_0_0
@@ -209,7 +210,6 @@ class WebFeatureService_(object):
                 if m.get("type").lower() == method.lower()
             )
         )
-        base_url = base_url if base_url.endswith("?") else base_url + "?"
 
         request = {"service": "WFS", "version": self.version, "request": "GetFeature"}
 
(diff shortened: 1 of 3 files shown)


justb4:
Test fails for same reason as with my PR #587 initially: 
```
tests/test_wfs3_pygeoapi.py::test_wfs3_pygeoapi FAILED
```
has nothing to do with your PR, but a recent change in `pygeoapi` and its demo server used in tests.

geographika:
Pull request rebased onto latest master branch and all tests passing. 

The coveralls check is failing - it reports -0.006% coverage change, but I don't think this is correct - the changes in this pull requests are tested as previously. It also reported -0.2% on another run with the same code. 

I ran into a similar issue on MapServer with small changes between runs. We increased the threshold to 1% to avoid this (see https://github.com/MapServer/MapServer/issues/6223). Maybe OWSLib is also required to have 60% coverage?

Issue #614, opened 2019-10-10
Exception calling TMS TileMapService constructor


justb4:
## Reproduce

* call `TileMapService(url, version='1.0.0')`
* failure
```
AttributeError: 'TileMapService' object has no attribute 'username'
```

## Context
In v0.18.0 and in GH `master`

## Analysis
See the following code in [tms.py#L41](https://github.com/geopython/OWSLib/blob/master/owslib/tms.py#L41):

```
    def __init__(self, url, version='1.0.0', xml=None, username=None, password=None,
                 parse_remote_metadata=False, timeout=30, auth=None):
        """Initialize."""
        if auth:
            if username:
                auth.username = username
            if password:
                auth.password = password
        self.url = url
        self.auth = auth or Authentication(username, password)
        self.version = version
        self.timeout = timeout
        self.services = None
        self._capabilities = None
        self.contents={}

        # Authentication handled by Reader                                                                                                                                                
        reader = TMSCapabilitiesReader(
===>                self.version, url=self.url, un=self.username, pw=self.password
                )
```

`self.username, pw=self.password` are never set.
## Solution
Assign  `self.username, pw=self.password` in constructor. Also used on [tms.py#L73](https://github.com/geopython/OWSLib/blob/master/owslib/tms.py#L73).

cehbrecht:
Fixed by PR #615.

PR #615, opened 2019-10-10
#614 fix broken TileMapService constructor

justb4:
Assign both `username` and `password` to `self` as these are referred to within `TMSCapabilitiesReader` and in `_getcapproperty()`.


cehbrecht:
@justb4 If it works I would accept. But we need a better solution as discussed in issue #609.

cehbrecht:
Fix for issue #614.

cehbrecht:
@justb4 merged. Thanks :)

PR #623, opened 2019-10-24
update OGC API - Features API description handling

tomkralidis:


tomkralidis:
Note that ldproxy WFS3 server needs to update to updates in the specification. Other errors are unrelated.

PR #628, opened 2019-10-31
Add required fields to wfs schema

Alexandre27:
I'm using the wfs get_schema method to replace geoserver's describeFeatureType endpoint and one thing that i found lacking for my use case is which fields can and cannot be null so I added a `"required"` entry to the existing schema, which has all fields that have `nillable="false"`.

I don't know if this is the right approach, so if anything needs to change, let me know.


owslib/feature/schema.py
```diff
@@ -124,13 +124,17 @@ def _construct_schema(elements, nsmap):
     for element in elements:
         data_type = element.attrib["type"].replace(gml_key + ":", "")
         name = element.attrib["name"]
+        non_nillable = element.attrib["nillable"] == "false"
```
Roel: I would replace this with something safer in the lines of
```python
non_nillable = element.attrib.get("nillable", "false") == "false"
```

On the basis that the `nillable` attribute on an xsd:element is optional, with a default of `false` as described in the [XSD schema](https://www.w3.org/2001/XMLSchema.xsd):
```xml
<xs:complexType name="element" abstract="true">
    <xs:annotation>
        <xs:documentation>
The element element can be used either at the top level to define an element-type binding globally, or within a content model to either reference a globally-defined element or type or declare an element-type binding locally. The ref form is not allowed at the top level.
        </xs:documentation>
    </xs:annotation>
    <xs:complexContent>
        <xs:extension base="xs:annotated">
            <xs:sequence>
                <xs:choice minOccurs="0">
                    <xs:element name="simpleType" type="xs:localSimpleType"/>
                    <xs:element name="complexType" type="xs:localComplexType"/>
                </xs:choice>
                <xs:group ref="xs:identityConstraint" minOccurs="0" maxOccurs="unbounded"/>
            </xs:sequence>
            <xs:attributeGroup ref="xs:defRef"/>
            <xs:attribute name="type" type="xs:QName"/>
            <xs:attribute name="substitutionGroup" type="xs:QName"/>
            <xs:attributeGroup ref="xs:occurs"/>
            <xs:attribute name="default" type="xs:string"/>
            <xs:attribute name="fixed" type="xs:string"/>
            <xs:attribute name="nillable" type="xs:boolean" use="optional" default="false"/>
            <xs:attribute name="abstract" type="xs:boolean" use="optional" default="false"/>
            <xs:attribute name="final" type="xs:derivationSet"/>
            <xs:attribute name="block" type="xs:blockSet"/>
            <xs:attribute name="form" type="xs:formChoice"/>
        </xs:extension>
    </xs:complexContent>
</xs:complexType>
```

cehbrecht:
@Roel @tomkralidis looks good to me. Save to merge?

tomkralidis:
+1 thanks

Roel:
Looks good to me too. We could include this in the tests here:
https://github.com/geopython/OWSLib/blob/master/tests/test_wfs_schema.py#L100
https://github.com/geopython/OWSLib/blob/master/tests/test_wfs_schema.py#L130

cehbrecht:
> Looks good to me too. We could include this in the tests here:
> https://github.com/geopython/OWSLib/blob/master/tests/test_wfs_schema.py#L100
> https://github.com/geopython/OWSLib/blob/master/tests/test_wfs_schema.py#L130

@Alexandre27 @Roel are you updating this PR? We can have it then in the 0.19.1 release.

Roel:
Since this is new functionality it does not per se need to be in 0.19.1 IMO.

@Alexandre27 can you extend the tests or shall I?

cehbrecht:
> Since this is new functionality it does not per se need to be in 0.19.1 IMO.

... let's use the current drive to get it in :)

Alexandre27:
@Roel I added tests, let me know if any change is needed.

Roel:
Thanks! Looks good to me :)

cehbrecht:
@Alexandre27 merged. Thanks :)

PR #634, opened 2019-12-19
Add python_requires to prevent pip installing this on py2

craigds:
owslib 0.19.0 successfully installs on unsupported pythons, and errors later because it only supports python 3. 

This PR adds a conservative `python_requires` for 3.4.

Usage:

```
$ pip2 install .
Processing /Users/craigds/checkout/OWSLib
ERROR: OWSLib requires Python '>=3.4' but the running Python is 2.7.16
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,6 +28,7 @@ setup(name              = 'OWSLib',
       maintainer_email  = 'tomkralidis@gmail.com',
       url               = 'https://geopython.github.io/OWSLib',
       install_requires  = reqs,
+      python_requires   = '>=3.5',
       cmdclass          = {'test': PyTest},
       packages          = find_packages(exclude=["docs", "etc", "examples", "tests"]),
       classifiers       = [


tomkralidis:
Where/how does OWSLib fail at run time?

craigds:
ugh, just read that after getting back from holiday...

I don't have the breakage with me at the moment, but... it's pretty clear that 2.x support has been dropped, right? see 430342c024f74d70a6cf78d1af810e803822997d

So the exact place it breaks shouldn't matter; the release should be marked as requiring python 3.something.

To me this seems like it'd be cause for yanking the current release from PyPI altogether, and uploading a new release with this PR merged. Other projects have done just that, e.g. just now pyhamcrest yanked their release for the same reason: https://github.com/hamcrest/PyHamcrest/issues/131

tomkralidis:
@craigds thanks for the info.  A trace of the breakage would have been valuable to help understand better the context.  Regardless, cc'ing @cehbrecht here for comments as well.  Having said this, should we require `>=3.5` per our [Travis](https://github.com/geopython/OWSLib/blob/master/.travis.yml#L2) setup?

cehbrecht:
@craigds @tomkralidis I would also require `>=3.5` as Tom said.

Just for reference the explanation of the `python_requires` parameter:
https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires

> @craigds thanks for the info. A trace of the breakage would have been valuable to help understand better the context. Regardless, cc'ing @cehbrecht here for comments as well. Having said this, should we require `>=3.5` per our [Travis](https://github.com/geopython/OWSLib/blob/master/.travis.yml#L2) setup?

craigds:
Okay, changed it to 3.5

PR #635, opened 2019-12-25
Py38

ocefpaf:
Added py38 to the test matrix and fixed `cgi.parse_qsl` deprecation.

Note that the current code is broken on Python 3.8 without these changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: python
 python:
   - "3.5"
   - "3.6"
+  - "3.8"
 # Enable 3.7 without globally enabling sudo and dist: xenial for other build jobs
 matrix:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,18 +2,8 @@ language: python
 python:
   - "3.5"
   - "3.6"
+  - "3.7"
   - "3.8"
-# Enable 3.7 without globally enabling sudo and dist: xenial for other build jobs
-matrix:
-  include:
-    - python: 3.7
-      dist: xenial
-      sudo: true
-      env: LMXL=true
-    - python: 3.7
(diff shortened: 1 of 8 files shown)


fmigneault:
@ocefpaf 
Would it be possible to have this tagged as `0.19.1` or similar to bump the version in pypi?
At the moment, anyone using Python 3.8 will automatically receive an error due to cgi deprecation when installing latest owslib.

cehbrecht:
@tomkralidis How about a 0.19.1 release? I can update the changes:
https://github.com/geopython/OWSLib/issues?q=is%3Aclosed+milestone%3A0.20.0

ocefpaf:
> @ocefpaf
> Would it be possible to have this tagged as `0.19.1` or similar to bump the version in pypi?

I don't have the proper rights here to do so but @cehbrecht is already on it.

> At the moment, anyone using Python 3.8 will automatically receive an error due to cgi deprecation when installing latest owslib.

If you are using the `conda-forge` version of the package we backported this change and you should be good to go.

fmigneault:
@cehbrecht Thanks for looking into this.
@ocefpaf I'm not using the conda package, using directly the pip requirements.
I'll wait for the version whenever it is available.

tomkralidis:
@cehbrecht sure, ping me once changes are updated and I will cut 0.19.1

cehbrecht:
@tomkralidis changes done ... just wait for PR #628.

cehbrecht:
@tomkralidis I think we are good to go for 0.19.1.

tomkralidis:
thanks @cehbrecht . 0.19.1 released.

fmigneault:
Thanks to all of you :)

Issue #636, opened 2020-01-02
WMS - xml to object conversion is None, with valid url


Leanderson-Coelho:
**Describe the bug**

When traversing the xml tags, the WebMapService_1_1_1 `/owslib/map/wms111.py` object's constructor cannot get the xml information, because for some strange reason the tags are pre-informed even before their name, as in Screenshot 1:

```
Tag: ContactOrganization
real value: {http://www.opengis.net/wms}ContactOrganization  
```

in xml: Screenshot 2

**To Reproduce**

> URL used: http://datageo.ambiente.sp.gov.br/serviceTranslator/rest/getXml/IPT_Geoserver_WMS/Risco_Louveira/1449249453587/wms

```
from owslib.wms import WebMapService

wms = WebMapService('URL used')
```
Exception thrown: Screenshot 3

This is because a tag with the name 'Service' was not found on line 95 of Screenshot 4. Soon it was sending None to method 'ServiceIdentification' line 96, which obviously does not have the method 'find ()' since it is a None.

Throwing the exception at line 358 Screenshot 5 because the object is None.

**Context**
OWSLib Version [latest on Dec 30, 2019]

**Screenshots or Logfiles**

Screenshot 1
![image](https://user-images.githubusercontent.com/34748960/71683086-4d054700-2d70-11ea-8d18-f79c8ce4cb1d.png)

Screenshot 2
![image](https://user-images.githubusercontent.com/34748960/71683199-9d7ca480-2d70-11ea-8e7b-af1338f58d40.png)

Screenshot 3
![exception](https://user-images.githubusercontent.com/34748960/71684274-cbafb380-2d73-11ea-8f91-8136952563b2.png)

Screenshot 4
![image](https://user-images.githubusercontent.com/34748960/71684379-1c271100-2d74-11ea-8494-571687547047.png)

Screenshot 5
![image](https://user-images.githubusercontent.com/34748960/71684541-8b046a00-2d74-11ea-96b6-ef1f913162c0.png)

PR #637, opened 2020-01-02
solution for issue #636

Leanderson-Coelho:
uma solução para issue#636 testada

### DIFF
--- a/owslib/map/wms111.py
+++ b/owslib/map/wms111.py
@@ -34,6 +34,7 @@ from owslib.namespaces import Namespaces
 
 n = Namespaces()
 
+
 class CapabilitiesError(Exception):
     pass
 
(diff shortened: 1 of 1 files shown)


cehbrecht:
@Leanderson-Coelho @tomkralidis 
I think this is not the right solution. It looks like your caps document is wms version 1.3.0 and you try to parse it with wms version 1.1.1. I tried it now with 1.3.0 and the parser was complaining about a missing `Title` element in the `Style`. 

I have added a test for this issue and it is available as PR #644. I can merge this PR and you could update it ... I don't know where a fix is needed, in owslib or on the service side. I'm not so familiar with the wms protocol.

cehbrecht:
@Leanderson-Coelho the test in PR #644 is merged. You can build on it and add a new PR.
 Should I close this one?

PR #644, opened 2020-01-27
test for issue 636

cehbrecht:
This PR adds a test for issue #636.


cehbrecht:
@tomkralidis thanks for quick review :)

PR #646, opened 2020-01-28
Remove redundant check on logging level

johanvdw:
log.debug will itself check the logging level. For one line logging it seems a bit redundant to do the check again.

### DIFF
--- a/owslib/coverage/wcs100.py
+++ b/owslib/coverage/wcs100.py
@@ -126,8 +126,7 @@ class WebCoverageService_1_0_0(WCSBase):
         except StopIteration:
             base_url = self.url
 
-        if log.isEnabledFor(logging.DEBUG):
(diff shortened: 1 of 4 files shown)


owslib/coverage/wcs100.py
```diff
@@ -126,8 +126,7 @@ def getCoverage(self, identifier=None, bbox=None, time=None, format=None, crs=No
         except StopIteration:
             base_url = self.url
 
-        if log.isEnabledFor(logging.DEBUG):
-            log.debug('WCS 1.0.0 DEBUG: base url of server: %s' % base_url)
+        log.debug('WCS 1.0.0 DEBUG: base url of server: %s' % base_url)
```
sbrunner: Should be:
`log.debug('WCS 1.0.0 DEBUG: base url of server: %s', base_url)`
=> Don't interpret the template if not needed...

PR #647, opened 2020-01-28
Fix WFS get_schema when typename doubles as attributename

### LINKED ISSUE #645: WFS get_schema broken in corner case where typename equals attribute name
In the corner case where the WFS typename also repeats as a column name, the WFS get_schema() method fails with the following error:
```python

hhz = WebFeatureService(
    'https://www.dov.vlaanderen.be/geoserver/wfs',
    version='1.1.0')

schema = hhz.get_schema('gw_varia:hhz')

print(schema)
```
```
  File "C:\Projecten\PyDov\OWSLib\owslib\feature\__init__.py", line 213, in get_schema
    return get_schema(self.url, typename, self.version, auth=self.auth)
  File "C:\Projecten\PyDov\OWSLib\owslib\feature\schema.py", line 62, in get_schema
    elements = _get_elements(complex_type, root)
  File "C:\Projecten\PyDov\OWSLib\owslib\feature\schema.py", line 79, in _get_elements
    )[0]
TypeError: 'NoneType' object is not subscriptable
```

This is because the wrong xsd:element is selected from the DescribeFeatureType XSD:
```xml
<xsd:schema elementFormDefault="qualified" targetNamespace="http://dov.vlaanderen.be/grondwater/gw_varia">
    <xsd:import namespace="http://www.opengis.net/gml" schemaLocation="https://www.dov.vlaanderen.be/geoserver/schemas/gml/3.1.1/base/gml.xsd"/>
    <xsd:complexType name="hhzType">
        <xsd:complexContent>
            <xsd:extension base="gml:AbstractFeatureType">
                <xsd:sequence>
                    <xsd:element maxOccurs="1" minOccurs="0" name="id" nillable="true" type="xsd:string"/>
                    <xsd:element maxOccurs="1" minOccurs="0" name="hhz_naam" nillable="true" type="xsd:string"/>
                    <xsd:element maxOccurs="1" minOccurs="0" name="opp_m2" nillable="true" type="xsd:long"/>
                    <xsd:element maxOccurs="1" minOccurs="0" name="hhz" nillable="true" type="xsd:string"/>
                    <xsd:element maxOccurs="1" minOccurs="0" name="geom" nillable="true" type="gml:GeometryPropertyType"/>
                </xsd:sequence>
            </xsd:extension>
        </xsd:complexContent>
    </xsd:complexType>
    <xsd:element name="hhz" substitutionGroup="gml:_Feature" type="gw_varia:hhzType"/>
</xsd:schema>
```

I plan to fix this in a PR in the next couple of days.

Roel:
The type_element we want is the one directly under the schema root. The first result from findall() where 'name="typename"' can give a wrong result when the typename doubles as the name of an attribute.

Includes tests for both the normal case as well as this corner case.

Fixes #645

### DIFF
--- a/owslib/feature/schema.py
+++ b/owslib/feature/schema.py
@@ -51,12 +51,7 @@ def get_schema(
 
     if ":" in typename:
         typename = typename.split(":")[1]
-    type_element = findall(
-        root,
-        "{%s}element" % XS_NAMESPACE,
-        attribute_name="name",
-        attribute_value=typename,
-    )[0]
+    type_element = root.find("./{%s}element" % XS_NAMESPACE)
     complex_type = type_element.attrib["type"].split(":")[1]
     elements = _get_elements(complex_type, root)
     nsmap = None
--- a/owslib/feature/schema.py
+++ b/owslib/feature/schema.py
@@ -46,8 +46,8 @@ def get_schema(
     else:
         auth = Authentication(username, password)
     url = _get_describefeaturetype_url(url, version, typename)
-    res = openURL(url, timeout=timeout, headers=headers, auth=auth)
-    root = etree.fromstring(res.read())
+    root = _get_remote_describefeaturetype(url, timeout=timeout,
+                                          headers=headers, auth=auth)
 
     if ":" in typename:
         typename = typename.split(":")[1]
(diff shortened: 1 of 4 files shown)


cehbrecht:
@Roel thanks :)

Issue #643, opened 2020-01-26
Docs are not up to date


pathmapper:
Hi,

according to the changelog WMS 1.3.0 support was added in `0.13.0`:

https://github.com/geopython/OWSLib/blob/master/CHANGES.rst#0130-2016-09-24

but it's not listed as supported: https://geopython.github.io/OWSLib/#standards-support

![grafik](https://user-images.githubusercontent.com/20856381/73140940-b2eaa300-407e-11ea-9673-182cab0268af.png)

pathmapper:
Also the readme of this repo gives more information than the docs page:

e.g. `wms.identification.version` and `wms.identification.abstract` are missing from the docs:

https://github.com/geopython/OWSLib:
![grafik](https://user-images.githubusercontent.com/20856381/73261739-f94e1800-41cc-11ea-9bda-a64071d0cd17.png)

https://geopython.github.io/OWSLib/:
![grafik](https://user-images.githubusercontent.com/20856381/73261651-cdcb2d80-41cc-11ea-948b-0231f397b97b.png)

cehbrecht:
@pathmapper done. Thanks for pointing at this :)

pathmapper:
Thanks @cehbrecht :+1:

PR #649, opened 2020-01-29
update wms docs

cehbrecht:
This PR fixes issue #643. It updates the wms docs as requested (list supported version 1.3.0)


cehbrecht:
@tomkralidis merged. just a minor doc change.

PR #651, opened 2020-01-30
Tests: use HEAD instead of GET in service_ok

Roel:
Since we only rely on the status code and the headers to determine if a service is working or not, it is not needed to perform a GET (resulting in downloads of data we don't need).

Instead we could use a HEAD request. Since Requests does not follow redirects by default when using HEAD [[1]], we enable it explicitly using allow_redirects.

[1]: https://2.python-requests.org/en/master/user/quickstart/#redirection-and-history

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -83,7 +83,7 @@ def sorted_url_query(url):
 
 def service_ok(url, timeout=5):
     try:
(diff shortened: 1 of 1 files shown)


cehbrecht:
Test failure not related to this PR.

cehbrecht:
@Roel merged. Thanks :)

PR #655, opened 2020-02-07
wps: parse Languages section in GetCapabilities

davidcaron:
An example of using the added `WebProcessingService.languages` attribute:

```python
>>> from owslib.wps import WebProcessingService
>>> wps = WebProcessingService(url='...')
>>> wps.language is None  # language is None when not specified (current behaviour)
True
>>> wps.languages.default  # check the default language
'en-US'
>>> wps.languages.supported  # check the supported languages
['en-US', 'fr-CA']
>>> wps.language = 'fr-CA'  # set from the supported languages list
```

### DIFF
--- a/owslib/wps.py
+++ b/owslib/wps.py
@@ -249,6 +249,7 @@ class WebProcessingService(object):
         self.provider = None
         self.operations = []
         self.processes = []
+        self.languages = None
 
         if not skip_caps:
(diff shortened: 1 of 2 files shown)


cehbrecht:

cehbrecht:
@davidcaron code looks nice and clean. Thanks for adding this :)

Issue #619, opened 2019-10-17
Error with WMTS example


dopplershift:
The code example from the docs:
```python
>>> from owslib.wmts import WebMapTileService
>>> wmts = WebMapTileService("http://map1c.vis.earthdata.nasa.gov/wmts-geo/wmts.cgi")
```
currently results in a traceback:
```pytb
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-11-d6e4fa9c2cd1> in <module>
      4 fig = plt.figure()
      5 ax = fig.add_subplot(1, 1, 1, projection=ccrs.PlateCarree())
----> 6 ax.add_wmts(url, layer)
      7 ax.set_extent([-15, 25, 35, 60], crs=ccrs.PlateCarree())
      8 

~/miniconda3/envs/py37/lib/python3.7/site-packages/cartopy/mpl/geoaxes.py in add_wmts(self, wmts, layer_name, wmts_kwargs, **kwargs)
   1939         from cartopy.io.ogc_clients import WMTSRasterSource
   1940         wmts = WMTSRasterSource(wmts, layer_name,
-> 1941                                 gettile_extra_kwargs=wmts_kwargs)
   1942         return self.add_raster(wmts, **kwargs)
   1943 

~/miniconda3/envs/py37/lib/python3.7/site-packages/cartopy/io/ogc_clients.py in __init__(self, wmts, layer_name, gettile_extra_kwargs)
    378                 hasattr(wmts, 'contents') and
    379                 hasattr(wmts, 'gettile')):
--> 380             wmts = owslib.wmts.WebMapTileService(wmts)
    381 
    382         try:

~/miniconda3/envs/py37/lib/python3.7/site-packages/owslib/wmts.py in __init__(self, url, version, xml, username, password, parse_remote_metadata, vendor_kwargs, auth)
    189 
    190         # build metadata objects
--> 191         self._buildMetadata(parse_remote_metadata)
    192 
    193     def _getcapproperty(self):

~/miniconda3/envs/py37/lib/python3.7/site-packages/owslib/wmts.py in _buildMetadata(self, parse_remote_metadata)
    239                     self.contents[cm.id] = cm
    240                 gather_layers(elem, cm)
--> 241         gather_layers(caps, None)
    242 
    243         self.tilematrixsets = {}

~/miniconda3/envs/py37/lib/python3.7/site-packages/owslib/wmts.py in gather_layers(parent_elem, parent_metadata)
[...]

PR #656, opened 2020-02-11
fix for cartopy/wmts examples, solves #619

raphaeldussin:
lot of examples out there making use of NASA's service (https://map1c.vis.earthdata.nasa.gov/wmts-geo/wmts.cgi) are broken. Based on chronology, I think it's related to changes in the WMTS service.
The problem OWSLib code is 5 years old and examples broke a few month ago.

Replacing the error by a warning solved the problem and gives the expected results.


cehbrecht:
@raphaeldussin @tomkralidis I'm not familiar with wmts ... but I would at least accept this as a workaround.

cehbrecht:

cehbrecht:
@raphaeldussin thanks :)

PR #657, opened 2020-02-11
Fix wms tests 20200211

cehbrecht:
This PR fixes the test suite:
* converted doctest  test_wmts_restonly
* fixed test_wms_getfeatureinfo_130


cehbrecht:
@tomkralidis just fixing broken tests.

PR #658, opened 2020-02-17
add WCS code example to documentation

misev:
in addition fix execution with python 2.

### DIFF
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -36,7 +36,7 @@ Standards Support
 +-----------------------+-----------------------------+
 | `OGC WFS`_            | 1.0.0, 1.1.0, 2.0.0, 3.0    |
 +-----------------------+-----------------------------+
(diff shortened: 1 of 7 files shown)


cehbrecht:
@misev Thanks for adding the wcs example. We have skipped Python 2.x support starting with release 0.19. Could you please revert the py2/py3 compatibility code?

misev:
I reverted the commit. The documentation should be updated here I guess to list pip3 instead of pip? http://geopython.github.io/OWSLib/#install

cehbrecht:
> The documentation should be updated here I guess to list pip3 instead of pip? http://geopython.github.io/OWSLib/#install

I usually work with Python 3 only in my environment ... so I don't need to distinguish. We could emphasis that it works only with Python 3. @tomkralidis what do you think?

cehbrecht:
@misev thanks :)

PR #659, opened 2020-02-20
drop last bits of python 2

ocefpaf:
This PR is a bit more complicated than it looks and you may delay the merge until OWSLib is ready to drop Python 3.5 support too.

The reason is that, with this PR, some responses are no longer converted to `str`. The main advantage is that everything is bytes in a way we expected from Python 3. The downside is that Python 3.5 `json` module cannot handle bytes, that started in 3.6, and some users may still expect `str` rather than bytes in their workflow. (See the docs test where I had to add the `b""` as an example of that.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
-  - "3.5"
   - "3.6"
   - "3.7"
   - "3.8"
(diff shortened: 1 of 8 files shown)


cehbrecht:
@tomkralidis I'm ok with dropping Python 3.5. What do you think?

cehbrecht:
@ocefpaf merged. Thanks :)

PR #661, opened 2020-02-20
Fix pyproj deprecation

ocefpaf:
In this PR:

- fixed the deprecated syntax in pyproj to avoid future breakages;
- be explicit about `always_xy=True` to prevent lat, lon vs lon, lat confusions;
- drop pyproj <2 to make it easier to support modern pyproj and proj without workaround.

If supporting pyproj 1.9.* is important we can add a few if-clauses here.


cehbrecht:
@tomkralidis looks good to me. Ok to merge?

tomkralidis:
+1

cehbrecht:
@ocefpaf merged. Thanks :)

Issue #662, opened 2020-02-24
param_list_to_url_string function in util.py function is not associated with any class but uses self parameter


njanakiev:
When running `getCoverage` in `owslib=0.19.1` for:

```python
geoserver_url = "..."
layer_name = "..."
lat_subset  = ('Lat',   40.47739,  40.916178)
long_subset = ('Long', -74.25909, -73.700181)

wcs = WebCoverageService(geoserver_url, version='2.0.1')

output = wcs.getCoverage(identifier=[layer_name],
    format='image/tiff', subsets=[long_subset, lat_subset])
```
I get the error: `TypeError: param_list_to_url_string() missing 1 required positional argument: 'param_name'` This snippet worked fine for `owslib=0.19.0`. 

I can see that in `utils.py`, the function `param_list_to_url_string(self, param_list, param_name)` uses `self` but is not associated to any class.

cehbrecht:
@njanakiev this looks like a bug. Would you like to make a PR?

Dukestep:
Sorry, I introduced this... I submitted a PR that rectifies this error (#663).

cehbrecht:
fixed by PR #663.

PR #663, opened 2020-02-25
fix param_list_to_url_string() util function and refactor

Dukestep:
The `params_list_to_url_string` in `owslib/util.py` contains an reference to `self` in its arguments and within the function itself. This was a remnant of when the function was a method of the `WebCoverageService_2_0_1` class during my initial implementation.

This PR removes this error and also moves some of the helper methods in the various WCS implementations into `owslib/util.py` in order to avoid code repetition between the different WCS versions.

See also #662.
 
cc @tomkralidis.

### DIFF
--- a/owslib/coverage/wcs100.py
+++ b/owslib/coverage/wcs100.py
@@ -18,7 +18,7 @@ import os
 import errno
 
 import logging
-from owslib.util import log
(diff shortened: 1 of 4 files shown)


cehbrecht:
@Dukestep code looks good. Thanks for the quick fix :)

PR #667, opened 2020-03-11
Fix reading style in getCap

mlefort:
When a style is found without "title" or "name", a ValueError is raised 

However, the style should only be ignored, otherwise it's too restrictive for some services
(ex : https://inspire.cadastre.gouv.fr/scpc/31555.wms?service=WMS&request=GetCapabilities)

### DIFF
--- a/owslib/map/wms130.py
+++ b/owslib/map/wms130.py
@@ -582,7 +582,7 @@ class ContentMetadata(AbstractContentMetadata):
             name = s.find(nspath('Name', WMS_NAMESPACE))
             title = s.find(nspath('Title', WMS_NAMESPACE))
             if name is None or title is None:
-                #raise ValueError('%s missing name or title' % (s,))
+                # raise ValueError('%s missing name or title' % (s,))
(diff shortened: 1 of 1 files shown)


cehbrecht:
@mlefort I can not say if this is a valid approach. Maybe some else familiar with WMS can comment? 

The tests are failing due to pep8 check:
```
$ flake8 owslib

owslib/map/wms130.py:585:17: E265 block comment should start with '# '
```

mlefort:
Thank you for your answer. I corrected the error linked to pep8 check. 
However, there seems to be always a problem with the tests on python 3.6, but I don't see the connection with my modification.

tomkralidis:
Note that `Name` and `Style` are required elements as per WMS 1.3.0.

mlefort:
You're right but that's not the point : if a style is incorrect, it should simply be ignored without generating an error. That's what other WMS clients do, such as QGIS. The lib should be able to handle WMS that is not perfect.

PR #670, opened 2020-03-18
fix getOutput

### LINKED ISSUE #669: WPS WPSExecution.getOutput bug
The code for `getOutput` does not seem to support downloading multiple outputs due to an indentation error.

cehbrecht:
This PR fixes #669. The `wps.getOutput` method write only a single complex output which can be specified with an optional `identifier` parameter.

### DIFF
--- a/owslib/wps.py
+++ b/owslib/wps.py
@@ -858,14 +858,23 @@ class WPSExecution(object):
 
         :param filepath: optional path to the output file, otherwise a file will be created in the local directory with
                   the name assigned by the server, or default name 'wps.out' for embedded output.
+        :oaram: identifier: optional identifier of the output that should be written.
+                  For backward compability if will default to first output.
         """
 
         if self.isSucceded():
             content = b''
-            for output in self.processOutputs:
-                if identifier and output.identifier != identifier:
-                    continue
-
+            output = None
+            if self.processOutputs:
+                if identifier:
+                    # filter outputs by identifier
+                    outputs = [o for o in self.processOutputs if o.identifier == identifier]
+                    if outputs:
+                        output = outputs[0]
+                else:
+                    # take the first found output
+                    output = self.processOutputs[0]
+            if output:
                 # ExecuteResponse contains reference to server-side output
                 if output.reference:
                     content = output.retrieveData(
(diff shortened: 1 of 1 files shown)


cehbrecht:
@huard can you review?

owslib/wps.py
```diff
@@ -851,7 +851,7 @@ def isSucceded(self):
     def isNotComplete(self):
         return not self.isComplete()
 
-    def getOutput(self, filepath=None):
+    def getOutput(self, filepath=None, identifier=None):
         """
         Method to write the outputs of a WPS process to a file:
         either retrieves the referenced files from the server, or writes out the content of response embedded output.
```
huard: document identifier parameter in docstring.

huard: document that it will default to first output for backward compatibility.

owslib/wps.py
```diff
@@ -863,34 +863,36 @@ def getOutput(self, filepath=None):
         if self.isSucceded():
             content = b''
             for output in self.processOutputs:
-
-                output_content = output.retrieveData(
-                    self.auth.username, self.auth.password,
-                    headers=self.headers, verify=self.auth.verify, cert=self.auth.cert)
+                if identifier and output.identifier != identifier:
```
huard: For clarity, set identifier explicitly to the first output. 
Use filter instead of continue. 
Remove break

owslib/wps.py
cehbrecht: done.

owslib/wps.py
cehbrecht: done.

cehbrecht:
@huard are you ok with the changes?

owslib/wps.py
```diff
@@ -851,46 +851,52 @@ def isSucceded(self):
...
 
-    def getOutput(self, filepath=None):
+    def getOutput(self, filepath=None, identifier=None):
         """
         Method to write the outputs of a WPS process to a file:
         either retrieves the referenced files from the server, or writes out the content of response embedded output.
 
         :param filepath: optional path to the output file, otherwise a file will be created in the local directory with
                   the name assigned by the server, or default name 'wps.out' for embedded output.
+        :oaram: identifier: optional identifier of the output that should be written.
```
huard:
```suggestion
        :param: identifier: optional identifier of the output that should be written.
```

owslib/wps.py
```diff
@@ -851,46 +851,52 @@ def isSucceded(self):
...
+                  For backward compability if will default to first output.
```
huard:
```suggestion
                  For backward compatibility it will default to the first output.
```

owslib/wps.py
cehbrecht: done.

owslib/wps.py
cehbrecht: done.

cehbrecht:
@huard done. thanks for review.

Issue #680, opened 2020-04-28
Support local files in wps.Output.retrieveData 


huard:
When testing WPS processes using the pywps TestClient, output files are stored in /tmp. It would be useful for testing if the owslib client could mock-download those files, as it would simplify writing tests for wps processes.  

We could then do something like:
```
client = client_for(Service(processes=[MyProcess()], cfgfiles=CFG_FILE))
ex = WPSExecution()

resp = client.get(service='wps', request='execute', version='1.0.0', identifier='my-process')
ex.parseResponse(resp.xml)
out = ex.processOutputs[0].retrieveData()
```

cehbrecht:
fixed by PR #681

PR #681, opened 2020-04-28
Add support in wps for retrieving data on local filesystem

huard:
Fix for #680 

Includes test.

### DIFF
--- a/owslib/wps.py
+++ b/owslib/wps.py
@@ -1430,6 +1430,9 @@ class Output(InputOutput):
                         headers=headers, verify=verify, cert=cert)
             # extract output filepath from URL query string
             self.fileName = spliturl[1].split('=')[1]
+        elif url.startswith("file://"):
+            u = open(url[7:])
+            self.fileName = url.split('/')[-1]
         else:
             u = openURL(
                 url, '', method='Get', username=username, password=password,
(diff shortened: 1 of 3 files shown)


owslib/wps.py
```diff
@@ -1430,6 +1430,9 @@ def retrieveData(self, username=None, password=None, headers=None, verify=True,
                         headers=headers, verify=verify, cert=cert)
             # extract output filepath from URL query string
             self.fileName = spliturl[1].split('=')[1]
+        elif url.startswith("file://"):
+            u = open(url[7:])
```
tomkralidis, line 1434: Would this result in an open file handle ?  Would adding a context manager help here (this block would need some logic rework) https://github.com/geopython/OWSLib/blob/c0d83476e91cf7ef0ec82134ffb828f95685ad06/owslib/wps.py#L1426-L1444)

huard: Not sure. Let's not take any chance.

cehbrecht:
@huard there is a common wps test:
https://github.com/geopython/OWSLib/blob/master/tests/test_wps.py

... and offline tests with predefined wps responses, like:
https://github.com/geopython/OWSLib/blob/master/tests/test_wps_execute.py

huard:
@cehbrecht  Test added.

cehbrecht:
@huard merged. Thanks :)

PR #685, opened 2020-05-11
Enable switching off SSL verification

Samweli:
Based on discussion from this issue https://github.com/geopython/OWSLib/issues/609

cc @gubuntu

### DIFF
--- a/owslib/util.py
+++ b/owslib/util.py
@@ -170,14 +170,14 @@ def openURL(url_base, data=None, method='Get', cookies=None, username=None, pass
             auth.password = password
         if cert:
             auth.cert = cert
-        if verify and not auth.verify:
-            auth.verify = verify
+        if not verify or not auth.verify:
+            verify = False
     else:
         auth = Authentication(username, password, cert, verify)
     if auth.username and auth.password:
         rkwargs['auth'] = (auth.username, auth.password)
     rkwargs['cert'] = auth.cert
-    rkwargs['verify'] = auth.verify
+    rkwargs['verify'] = verify
 
     # FIXUP for WFS in particular, remove xml style namespace
     # @TODO does this belong here?
--- a/owslib/util.py
+++ b/owslib/util.py
@@ -170,8 +170,7 @@ def openURL(url_base, data=None, method='Get', cookies=None, username=None, pass
             auth.password = password
         if cert:
             auth.cert = cert
-        if not verify or not auth.verify:
-            verify = False
+        verify = verify and auth.verify
     else:
         auth = Authentication(username, password, cert, verify)
     if auth.username and auth.password:


owslib/util.py
```diff
@@ -170,14 +170,14 @@ def openURL(url_base, data=None, method='Get', cookies=None, username=None, pass
             auth.password = password
         if cert:
             auth.cert = cert
-        if verify and not auth.verify:
-            auth.verify = verify
+        if not verify or not auth.verify:
+            verify = False
```
sbrunner, line 174: For me, this look strange, it can be simplified to 
```suggestion
        verify = verify and auth.verify
```

Samweli: right @sbrunner!

cehbrecht:
the code in openurl is confusing and we need to refactor it later. There is also a proposed workaround mentioned in the chat:
https://gitter.im/geopython/OWSLib

tomkralidis:
@Samweli thanks for the PR.  So the basic approach here is that unless verification is passed in BOTH `verify=True` and `auth.verify=True`, verification is turned OFF, correct?

Agree with @cehbrecht this is a workaround for the time being.  I'll add some documentation and notes in our release as well (this fix will trigger a new release).

Samweli:
> @Samweli thanks for the PR. So the basic approach here is that unless verification is passed in BOTH `verify=True` and `auth.verify=True`, verification is turned OFF, correct?
> 
> Agree with @cehbrecht this is a workaround for the time being. I'll add some documentation and notes in our release as well (this fix will trigger a new release).

Yes correct @tomkralidis, thanks.

tomkralidis:
@Samweli FYI the fix is not quite working for me:
```python
from owslib.csw import CatalogueServiceWeb as csw
from owslib.util import Authentication

auth = Authentication()

c = csw('https://geocris2.cdema.org/catalog/', auth=auth)
print(c.response)
```
Results in `OpenSSL.SSL.Error: [('SSL routines', 'tls_process_server_certificate', 'certificate verify failed’)]`

Samweli:
> @Samweli FYI the fix is not quite working for me:

Hi @tomkralidis, seems the package is not updated, it works if I install the package from source  and I initialize Authentication object with verify to False because otherwise it defaults to True.

```python

from owslib.csw import CatalogueServiceWeb as csw
from owslib.util import Authentication

auth = Authentication(verify=False)

c = csw('https://geocris2.cdema.org/catalog/', auth=auth)
print(c.response)
```

tomkralidis:
Hi @Samweli , to clarify, a client must ONLY pass `verify=False` to turn off SSL verification, correct?

Samweli:
> Hi @Samweli , to clarify, a client must ONLY pass `verify=False` to turn off SSL verification, correct?

Yes, correct @tomkralidis

PR #686, opened 2020-05-13
Add possibility to specify headers for WCS requests

TimoSkyGeo:
Dear maintainer,

I was missing the ability to specify the headers for a WCS request. 
Therefore, I made the necessary modifications to make this work.
Are there other actions I can take that would increase the likelihood that this functionality becomes part of your codebase?

Cheers,
Timo Millenaar


TimoSkyGeo:
I closed the pull request for after some further investigation I noticed I might as well send a request using the requests library and I found I did not properly take all versions into account, only 2.0.1. If you thing this is still interesting for you I am willing to spend some more effort on this. For now, this is closed.

PR #689, opened 2020-05-26
remove wfs test for version 1.0.0

cehbrecht:
Tests on travis CI are currently hanging due to wfs schema test with version 1.0.0. As a quick-fix I have removed this version from the test.


cehbrecht:
ping @Alexandre27

PR #695, opened 2020-06-08
add python version metadata

fmigneault:
adds some metadata that some parsers use to detect package supported python versions


cehbrecht:
@fmigneault thanks :)

Issue #683, opened 2020-05-09
POST WFS getfeatures 1.1.0 using filters broken? 


gilav:
Hello

I try to query a WFS server by data range + bbox, using POST method:
...
aBbox = BBox([pp1.x,pp1.y, pp2.x, pp2.y])
# date range
filter1 = PropertyIsGreaterThanOrEqualTo("imaging_time", "2020-01-24T16:46:41.616Z")
filter2 = PropertyIsLessThanOrEqualTo("imaging_time", "2020-01-27T16:46:41.616Z")
filterList = [filter1,filter2, aBbox]
#
fr1  = FilterRequest()
filter_fes = fr1.setConstraintList(filterList, tostring=True)
response = wfs11.getfeature(typename='GeoCenter:rsat2Features', filter=filter_fes, method="post")
...

Querying by bbox only using GETmethod works well.
Look like the issue is that the post data are not well formatted, It is urlEncoded, just like the GET querystring.

I suppose the post data has to ba valid xml, right?

Is there a workarround?

best regards

Gilles

Issue #439, opened 2018-04-10
WFS GetFeature via POST


Roel:
Is there support for sending WFS GetFeature requests via POST? I'd like to query the WFS service using a prepared WFS GetFeature filter in XML, but I can't get it to work with the current version.

The 'method' parameter seems to suggest it should work:
https://github.com/geopython/OWSLib/blob/d2a9722685de09e4f70189cdc034a23f5ecccffa/owslib/feature/wfs110.py#L154-L157
But changing this to `method='Post'` and providing a filter as XML doesn't seem to work.

I think the culprit is in the
https://github.com/geopython/OWSLib/blob/d2a9722685de09e4f70189cdc034a23f5ecccffa/owslib/feature/wfs110.py#L247

I can prepare a PR to support a (fourth) way of using the GetFeature method, with just a filter (in XML) and method = 'Post'? Like so:
```
wfs = WebFeatureService(
                url="https://www.dov.vlaanderen.be/geoserver/wfs",
                version="1.1.0")

filter = """<wfs:GetFeature service="WFS" version="1.1.0"
  xmlns:wfs="http://www.opengis.net/wfs"
  xmlns:ogc="http://www.opengis.net/ogc"
  xmlns:gml="http://www.opengis.net/gml"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://www.opengis.net/wfs
                      http://schemas.opengis.net/wfs/1.1.0/wfs.xsd">
  <wfs:Query typeName="gw_meetnetten:meetnetten">
    <ogc:Filter>
      <ogc:And>
      <ogc:Within>
        <ogc:PropertyName>geom</ogc:PropertyName>
[...]

PR #706, opened 2020-09-23
Add POST support to WFS GetFeature (Issue #439)

f-PLT:
This PR would take care of issues #439 and possibly #683 

This is my first contribution to OWSLib and I'm new to the WFS standard, so I expect there might be a few things to change, either towards the standards or to integrate it better with existing code.

I added the ability to use an existing xml filter as is, like suggested in #439, but I think it might be redundant. Either way, it's self contained and only 2 lines of code, so it's easy to remove if need be.

I can already say I'm not sure about how I handled Stored Queries.

### DIFF
--- a/owslib/feature/__init__.py
+++ b/owslib/feature/__init__.py
@@ -299,11 +299,6 @@ class WebFeatureService_(object):
         4) filter only via Post method
         """
 
-        if storedQueryID:
-            log.warning("Use of the storedQueryID argument is not yet implemented with the Post method")
-        if storedQueryParams:
-            log.warning("Use of the storedQueryParams argument is not yet implemented with the Post method")
-
         try:
             base_url = next(
                 (
@@ -318,13 +313,20 @@ class WebFeatureService_(object):
         if not typename and filter:
             return base_url, filter
 
+        request = self.create_post_request()
+
+        if storedQueryID:
+            storedQueryParams = storedQueryParams or {}
+            request.create_storedquery(storedQueryID, storedQueryParams)
+            data = request.to_string()
+            return base_url, data
+
         if typename:
             typename = (
                 [typename] if isinstance(typename, str) else typename
             )  # noqa: E721
             typenames = ",".join(typename)
 
-        request = self.create_post_request()
         request.create_query(typenames)
 
         if featureid:
@@ -343,8 +345,6 @@ class WebFeatureService_(object):
             request.set_maxfeatures(maxfeatures)
         if outputFormat:
             request.set_outputformat(outputFormat)
-        if startindex:
-            request.set_startindex(startindex)
         if propertyname:
             propertyname = (
                 [propertyname] if isinstance(propertyname, str) else propertyname
@@ -355,6 +355,8 @@ class WebFeatureService_(object):
                 [sortby] if isinstance(sortby, str) else sortby
             )
             request.set_sortby(sortby)
+        if startindex:
+            request.set_startindex(startindex)
 
         data = request.to_string()
         return base_url, data
--- a/owslib/feature/__init__.py
+++ b/owslib/feature/__init__.py
@@ -5,12 +5,15 @@
 #
 # =============================================================================
 
-from owslib.crs import Crs
-
-from urllib.parse import urlencode
 import logging
+from urllib.parse import urlencode
+from owslib.crs import Crs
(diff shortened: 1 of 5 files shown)


huard:
I'm not familiar with this part of the code, but the use case for us to send requests to GeoServer that exceed the size of GET requests. For example, we're sometime making queries for properties for thousands of individual features, and the list of features does not fit in GET requests. So we'd be really grateful if this could be reviewed.

cehbrecht (commented):
The code looks fine .... but I have not tested.

cehbrecht:
test failures not related to this PR.

KorkXx:
Just added #711 which this PR should solve

owslib/feature/__init__.py
```diff
@@ -206,6 +248,131 @@ def getGETGetFeatureRequest(
...
+                    if m.get("type").lower() == method.lower()
+                )
+            )
+        except StopIteration:
+            base_url = self.url
+
+        if not typename and filter:
+            return base_url, filter
+
+        request = self.create_post_request()
```
KorkXx, line 325: I did not understand why using a create_post_request function instead of directly inserting the version-dependent Postrequest object creation. Is this function intended to be used anywhere else ?

f-PLT: Well, my aim here was to make the post request creation more testable. It being separated makes it testable by itself.

I chose to make a factory with create_post_request(), so `getPOSTGetFeatureRequest()` can be used for wfs1.1.0 and 2.0.0 without needing another function that does almost the same thing but not quite, which also seems to be what `getGetGetFeatureRequest()` is aiming to become.

So this way, getFeature(), whether it's with 1.1.0 or 2.0.0, calls the same function, which checks the WFS version, and builds the appropriate request format. It will also make it easier (I hope) to integrate the next WFS version by simply extending the `postrequest.py` module so it can format post requests for that new version. All without needing to change `getPOSTGetFeatureRequest()` or creating a new function.

On the plus side, `postrequest.py` can also be used by itself to build a request and use it with resquests.post(), which could be useful in identifying if an error is caused by the request itself or if the problem is somewhere else in OWSLib.

cehbrecht:
@KorkXx @f-PLT @tomkralidis If the review is ok, can we merge it?

tomkralidis:
+1

f-PLT:
We have an issue with a usecase for one of our partners.

If it can't be handled server-side, we'll probably have a small change to add so it's possible to handle an https redirection, which doesn't seem to work with Post requests.

Right now, considering a simple flag to prevent the GetCapabilities url retrieval when needed, and force the use of the provided WebFeatureService url as is. (for reference : https://github.com/f-PLT/OWSLib/blob/wfs-getfeature-post/owslib/feature/__init__.py#L312-L320)

f-PLT:
Alright, it's definitely a server-side problem so we'll fix it server-side. If there are no other objections after the testing has been done , it should be ok to merge.

As for the problem, here's a summary as it's bound to happen to other people.

Post requests are not redirected. Ever, and should never be. It's in the HTTP standards. (Learn something everyday!)

Therefore if the the Url listed in the GetCapabilities for post getFeature() is `http://...`, but the server is behind, say, an nginx proxy that redirects `http => httpS`, the post request will not work and return a very helpful error : 
```
owslib.util.ServiceException: Could not determine geoserver request from http request org.geoserver...
```
The server needs to be configured to give the proper url. In the case of Geoserver, it should be the `Proxy Base Url`.

This doesn't happen with Get requests, because a Get can be redirected without any problem.

Hope this little bit of information can help someone someday, and if I'm wrong in any way, please feel free to correct me!

cehbrecht:
> Alright, it's definitely a server-side problem so we'll fix it server-side. If there are no other objections after the testing has been done , it should be ok to merge.

@f-PLT  ok :) Give me a sign ... I can do the merge ...

f-PLT:
Our tests have gone well!

Found and corrected a small unintended side effect while we were at it. If those last two commits pass muster, then it's good to go. If we end up finding more bugs, I'll open a new PR.

cehbrecht (approved):

cehbrecht:
failing tests are not related to this PR ... they already fail on current master.

cehbrecht:
@f-PLT merged. Thanks 👍

PR #707, opened 2020-09-25
Fix empty default value

ruz76:


owslib/util.py
```diff
@@ -591,6 +591,10 @@ def dump(obj, prefix=''):
 def getTypedValue(data_type, value):
     '''Utility function to cast a string value to the appropriate XSD type. '''
 
+    # If the default value is empty
+    if value is None:
+        return value
```
sbrunner, line 596:
```suggestion
        return
```

ruz76: OK for me. Thanks.

ruz76: The build failed, but the error is not connected to this pull request.

owslib/util.py
```diff
@@ -591,6 +591,10 @@ def dump(obj, prefix=''):
 def getTypedValue(data_type, value):
     '''Utility function to cast a string value to the appropriate XSD type. '''
 
+    # If the default value is empty
+    if value is None:
+        return
```
cehbrecht, line 596: I would expect that the function returns `None`?

ruz76: The original code acts in a different way. If the value is None and type for example integer it tries to cast None into integer and the code fails.

def getTypedValue(data_type, value):
    '''Utility function to cast a string value to the appropriate XSD type. '''

    if data_type == 'boolean':
        return True if value.lower() == 'true' else False
    elif data_type == 'integer':
        return int(value)
    elif data_type == 'float':
        return float(value)
    elif data_type == 'string':
        return str(value)
    else:
        return value  # no type casting 

So the correction may be either proposed or corrected by @sbrunner. The both returns None, but corrected has less code.

if value is None:
        return

PR #712, opened 2020-11-17
Allow requests' AuthBase derived authenticators to be used with OWSLib

achapkowski:
This is a simple PR to allow for Requests Authentication handlers to be given to the URL requests.

- allow AuthBase to be given

### DIFF
--- a/owslib/util.py
+++ b/owslib/util.py
@@ -177,7 +177,11 @@ def openURL(url_base, data=None, method='Get', cookies=None, username=None, pass
         auth = Authentication(username, password, cert, verify)
     if auth.username and auth.password:
         rkwargs['auth'] = (auth.username, auth.password)
-    rkwargs['cert'] = auth.cert
+    if hasattr(auth, 'cert'):
+        rkwargs['cert'] = auth.cert
+    elif cert:
+        rkwargs['cert'] = cert
+
     rkwargs['verify'] = verify
 
     # FIXUP for WFS in particular, remove xml style namespace
--- a/owslib/util.py
+++ b/owslib/util.py
@@ -24,6 +24,7 @@ import re
 from copy import deepcopy
 import warnings
 import requests
+from requests.auth import AuthBase
 import codecs
 
 """
@@ -162,8 +163,9 @@ def openURL(url_base, data=None, method='Get', cookies=None, username=None, pass
     rkwargs = {}
 
     rkwargs['timeout'] = timeout
-
-    if auth:
+    if isinstance(auth, AuthBase):
+        rkwargs['auth'] = auth
+    elif auth:
         if username:
             auth.username = username
         if password:


cehbrecht (approved):

cehbrecht:
@tomkralidis ok to merge?

tomkralidis:
LGTM.  Also cc'ing @justb4 for any comments.

justb4 (changes requested):
I find the added code somewhat confusing and could contain a potential error. I think passing an auth object is powerful. But now the `auth` can also be a `requests` AuthBase-derived object. E.g. later on there is an assignment `rkwargs['cert'] = auth.cert`. In case the `auth` object is AuthBase-derived, would that not cause an Exception? Same for `auth.username`. 

In the long term I would prefer requests' `AuthBase` derived, but realize there is lots of existing code, still passing username/password etc.

Also the function-comments section does not reflect that auth could be of type `AuthBase`.

achapkowski:
@justb4 can you post where you see the flaw?  auth.cert = False should be set on the request.<Method>

achapkowski:
Down the road, you should probably switch from requests.<METHOD> to just using a Session object.  I was looking at the LOE, but figured this was a good first go.

achapkowski:
I added some additional checks.  @justb4

cehbrecht:
@justb4 comments? good to merge?

justb4:
Hard to tell without extensive testing. This code has many possible scenario's based on what is passed to `openURL()`. For one thing the `auth` parameter needs additional comment, that it is either None, an OWSLib `Authentication` or an `requests AuthBase` derived object. `openURL()` is also called by many services. There it is always assumed that the `auth` object is of type OWSLib `Authentication`, at least containing a `username` and `password`. If a `requests AuthBase` derived object is passed those fields may not be present, for example a token auth. At least check with `hasattr()` then. Sorry, maybe someone else can have a fresh look. I am not too deep into OWSLib, but implemented several Auth-client systems in GeoHealthCheck with `requests`, including ESRI Token. So the essential code of this PR is:

```
    if isinstance(auth, AuthBase):
        rkwargs['auth'] = auth
    elif auth:
        if username:
            auth.username = username
        if password:
            auth.password = password
        if cert:
            auth.cert = cert
        verify = verify and auth.verify
    else:
        auth = Authentication(username, password, cert, verify)
    if auth.username and auth.password:
        rkwargs['auth'] = (auth.username, auth.password)
    if hasattr(auth, 'cert'):
        rkwargs['cert'] = auth.cert
    elif cert:
        rkwargs['cert'] = cert
```

It is the above code that confuses me: if a `requests AuthBase` derived object is passed (the `elif`) I would expect it to be directly passed in `rkwargs`. Now it is overwritten with username/password, and rkwargs['auth'] gets `auth.username, auth.password` tuple. [requests API](https://requests.readthedocs.io/en/latest/api/) expects *auth – (optional) Auth tuple or callable to enable Basic/Digest/Custom HTTP Auth.* Or am I completely missing something?

achapkowski:
The auth doc inherits from `AuthBase` which allows for customized authentication handlers.  So you can have something like a username/password authenticator that handles the token generation, NTLM, Kerberos, etc...  To check if the auth is of the right type, it should be an instance of `AuthBase`.

justb4:
Yes, but my point is that `if auth.username and auth.password:` . If `auth` is an `AuthBase` instance those attributes may not exist within the `auth` object. So at least there presence should be checked with `hasattr`. But still I find it strange that `auth` is not directly assigned to `rkwargs` and dus passed unaltered to `requests`. 

Again: I may be missing something, so anyone can have a fresh look? @cehbrecht maybe?

symbioquine:
@achapkowski thanks for working on this!

I also need this feature, and agree the changes proposed in this PR are hard to reason about because there are many places in OWSLib that assume the `auth` argument is an instance of the `owslib.util.Authentication` class...

@justb4 / @cehbrecht I've raised a [PR for an alternate implementation](https://github.com/geopython/OWSLib/pull/724) that I think may be easier to reason about. Looking forward to your thoughts!

achapkowski:
@symbioquine thanks for taking a deeper dive.  I don't use the whole API, so I might not have tested everything.  Mainly WMST, so if you feel there is a greater coverage, we can use your PR.  I think the `requests` authenticators make life easier, so if we can leverage them, let's do that!

justb4:
@symbioquine yes, LGTM! So my proposal is to follow [your PR](https://github.com/geopython/OWSLib/pull/724).

achapkowski:
Sounds good, closing

Issue #713, opened 2020-11-19
No way for user to adjust timeout limit on WCS1 getcoverage requests.


SpacemanPaul:
Probably applies to other services and versions as well, but as an example: 

WebCoverageService_1_0_0.getCoverage() calls OpenURL with no timeout parameter, so the timeout is always the default of 30 seconds.

It would be handy if you could specify a timeout parameter that would be passed through to OpenURL.

SpacemanPaul:
Have a quick dirty pull request.

PR #714, opened 2020-11-19
Allow user to specify timeout on WCS GetCoverage calls.

SpacemanPaul:
Add timeout parameter to getCoverage() methods for all WCS versions and pass through to underlying OpenUrl function. See #713


cehbrecht (approved):

cehbrecht:
@SpacemanPaul merged. Thanks :)

PR #719, opened 2020-12-02
fix wps bbox

cehbrecht:
The PR fixes the wps bbox parsing of the execute response. See PR in pywps:
https://github.com/geopython/pywps/pull/552


tomkralidis (approved):

cehbrecht:

PR #720, opened 2020-12-03
Fix tests

cehbrecht:
This PR is fixing the test suite.

Changes:
* test python 3.9 on travis
* fixed url in wfs tests
* fixed wmts test
* replaced deprecated `getiterator()` method from `xml.etree` for Python 3.9:
https://docs.python.org/3.8/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.getiterator

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ python:
   - "3.6"
   - "3.7"
   - "3.8"
+  - "3.9"
 sudo: false
 env:
   - LXML=true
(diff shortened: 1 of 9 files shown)


cehbrecht:
@Roel It looks like the pytest `monkeypatch` for `owslib.util.openURL` is not working as expected:
https://github.com/geopython/OWSLib/blob/af85567a8dc4ffa22d56298e928ce66fd2476f86/tests/test_remote_metadata.py#L207

cehbrecht:
@tomkralidis ok to merge?

tomkralidis:
+1 @cehbrecht , thanks

PR #724, opened 2020-12-21
Support requests' AuthBase in addition to basic password authentication

symbioquine:
* Why? OWSLib should support more kinds of authentication
  OAuth2/Digest/etc without needing to change further.

* How? Since the `Authentication` object is already plumbed throughout
  OWSLib, just add a new `auth_delegate` property which can be any object
  derived from `requests.AuthBase` and make sure that is honored at the
  points where actual requests' kwargs are built.

### DIFF
--- a/owslib/util.py
+++ b/owslib/util.py
@@ -24,6 +24,7 @@ import re
 from copy import deepcopy
 import warnings
 import requests
+from requests.auth import AuthBase
 import codecs
 
 """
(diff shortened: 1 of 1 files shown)


achapkowski:
The issues seem to fail on a service timeout, I think this PR is good.

cehbrecht (approved):

PR #727, opened 2021-01-04
migrate to GitHub actions

tomkralidis:


kalxas (approved):

.github/workflows/main.yml
```diff
@@ -0,0 +1,34 @@
...
+        LXML: ${{ matrix.lxml }}
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+      name: Setup Python ${{ matrix.python-version }}
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install requirements 📦
+      run: |
+        python3 -m pip install --upgrade pip
```
sbrunner, line 22: Is it relay needed?

sbrunner (approved):

.github/workflows/main.yml
kalxas: I agree we can drop

cehbrecht (approved):

PR #730, opened 2021-01-11
Add timeout to get capabilities

### LINKED ISSUE #728: Timeout for GetCapabilities (WMS, WMTS, WFS)
For WebMapService, WebFeatureService, WebMapTileService, etc... We can specify a timeout. This timeout is not passed to the reader (ex here: https://github.com/geopython/OWSLib/blob/master/owslib/wmts.py#L173 or https://github.com/geopython/OWSLib/blob/87e0bf83d60fbbc22903cb2056c198fb34ddc58b/owslib/map/wms130.py#L70 or https://github.com/geopython/OWSLib/blob/87e0bf83d60fbbc22903cb2056c198fb34ddc58b/owslib/feature/wfs100.py#L139)

And the default value in [openUrl](https://github.com/geopython/OWSLib/blob/87e0bf83d60fbbc22903cb2056c198fb34ddc58b/owslib/util.py#L145) is 30 seconds. Which is not enough in some rare case.

Maybe we could pass the timeout parameter to the reader class? I think creating a specific parameter for this (like `get_capabilities_timeout`) is  not necessary.

RemiDesgrange:
Fix #728 

Timeout were not taken into account for get capabilities. Default was 30s, and could not be customized.

I've modified the code to pass the timeout to the `*CapabilitiesReader` (`WFSCapabilitiesReader`, etc...). for WCS, WFS and WMS.

### DIFF
--- a/owslib/coverage/wcs100.py
+++ b/owslib/coverage/wcs100.py
@@ -38,17 +38,18 @@ class WebCoverageService_1_0_0(WCSBase):
         else:
             raise KeyError("No content named %s" % name)
 
-    def __init__(self, url, xml, cookies, auth=None):
+    def __init__(self, url, xml, cookies, auth=None, timeout=30):
         super(WebCoverageService_1_0_0, self).__init__(auth)
         self.version = '1.0.0'
         self.url = url
         self.cookies = cookies
+        self.timeout = timeout
         # initialize from saved capability document or access the server
-        reader = WCSCapabilitiesReader(self.version, self.cookies, self.auth, timeout=timeout)
+        reader = WCSCapabilitiesReader(self.version, self.cookies, self.auth)
         if xml:
             self._capabilities = reader.readString(xml)
         else:
-            self._capabilities = reader.read(self.url)
+            self._capabilities = reader.read(self.url, self.timeout)
 
         # check for exceptions
         se = self._capabilities.find('ServiceException')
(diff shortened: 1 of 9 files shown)


cehbrecht (changes requested):
Tests are failing:
```
got an unexpected keyword argument 'timeout'
```

RemiDesgrange:
@cehbrecht Thanks for your review. I now have trouble with coverall not launching but maybe not related to what I modified :thinking: ?

cehbrecht (approved):

cehbrecht:
@RemiDesgrange Looks good to me. Thanks :) Test failures are not related. I try to fix them in a new PR.

owslib/coverage/wcs100.py
```diff
@@ -38,17 +38,18 @@ def __getitem__(self, name):
         else:
             raise KeyError("No content named %s" % name)
 
-    def __init__(self, url, xml, cookies, auth=None):
+    def __init__(self, url, xml, cookies, auth=None, timeout=30):
         super(WebCoverageService_1_0_0, self).__init__(auth)
         self.version = '1.0.0'
         self.url = url
         self.cookies = cookies
+        self.timeout = timeout
```
tlvu, line 46: @RemiDesgrange I think there is a bug here.  You are setting `self.timeout = timeout` for `wcs100.py` only but not for all the other files.

This might explain why we are getting `AttributeError: 'WebCoverageService_2_0_1' object has no attribute 'timeout'` in https://github.com/Ouranosinc/raven/pull/348#issuecomment-770118401.

PR #736, opened 2021-01-25
OWS/ISO metadata updates

tomkralidis:
Minor safeguarding/additions:

- add OWS service metadata keyword type
- safeguard ISO thesaurus type


tomkralidis:
Travis failures unrelated.

Issue #737, opened 2021-01-27
WCS: creating instance fails with AttributeError


MartinPontius:
Hi,

I use owslib version 0.22.0.

When creating a new WCS instance

```
wcs = WebCoverageService(url='http://nrt.cmems-du.eu/motu-web/wcs', 
                               version='2.0.1', 
                               auth=auth)
```

the code fails with

`AttributeError: 'WebCoverageService_2_0_1' object has no attribute 'timeout'`

Error is thrown in line 67 in wcs201.py 

` self._capabilities = reader.read(self.url, self.timeout)`

This line was changed in this merge recently: https://github.com/geopython/OWSLib/pull/730

wildintellect:
I can confirm this bug was introduced in 0.22.0, code was working fine on 0.21.0
I think the issue is that init sets a default but the var name is timeout, it never gets assigned as part of the self object.
Fix should be adding `self.timeout = timeout` in the `_init_` of the class OR to use `timeout` not self.timeout 
`self._capabilities = reader.read(self.url, timeout)`

tlvu:
This issue is most probably fixed by PR https://github.com/geopython/OWSLib/pull/738

MartinPontius:
Yes, PR [738](https://github.com/geopython/OWSLib/pull/738) fixes the bug. The new release 0.23.0 is working again.

PR #738, opened 2021-01-30
Fix bug with timeout not assigned

RemiDesgrange:
Following this discussion https://github.com/geopython/OWSLib/pull/730#discussion_r567188656 I fixed the bug regarding timeout var not assigned.

### DIFF
--- a/owslib/coverage/wcs110.py
+++ b/owslib/coverage/wcs110.py
@@ -56,6 +56,7 @@ class WebCoverageService_1_1_0(WCSBase):
 
         self.url = url
         self.cookies = cookies
+        self.timeout = timeout
         # initialize from saved capability document or access the server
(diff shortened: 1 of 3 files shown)


cehbrecht:
test failures not related.

cehbrecht:
@RemiDesgrange merged. Thanks 😃

tlvu:
@cehbrecht If you can create a owslib release with this fix, it would be great.  It seems other than us is broken and need this fix, see https://github.com/geopython/OWSLib/issues/737

cehbrecht:
> @cehbrecht If you can create a owslib release with this fix, it would be great. It seems other than us is broken and need this fix, see #737

@tomkralidis Would you like to make a patch release?

PR #748, opened 2021-02-17
Add json responses from the 52N OGCAPI test server implementation

huard:
Could be used for static testing.


cehbrecht:
@huard adding a readme with the requests (urls)? you can merge it to:
https://github.com/geopython/OWSLib/tree/feature-749-ogcapi-processes

huard:
Done, but git doesn't allow me to push.

cehbrecht:
> Done, but git doesn't allow me to push.

you mean you can not merge? I think should have rights ... you can change the base branch from master to `feature-749-ogcapi-processes`

fmigneault:
The first step of `API` in https://github.com/geopython/OWSLib/blob/master/owslib/ogcapi is to fetch the JSON or YAML schema of the remote API. So I think evaluated instances should use the response they return for validation. 

For static analysis, maybe we should target the official https://github.com/opengeospatial/ogcapi-processes/ schemas ?

huard:
@cehbrecht I can't push to your branch... 

@fmigneault I'm not sure I know how we would test the client against the schemas. What do you have in mind ?

cehbrecht (approved):

cehbrecht:
@huard merged. Can also use issue #749 for discussion.

fmigneault:
@huard 
I just mean that instead of using static files posted to this repo which can become out of sync, we should use the ogc-api references directly. 
Static testing of any endpoint could then be evaluated against against those schemas to detect where API responses do not respect the reference OpenAPI. 
That would allow owslib users to run conformance tests against official schemas, which is something notably lacking for WPS.

fmigneault:
@huard 
Using something similar to https://pypi.org/project/openapi-schema-validator/
Compare the received JSON response with expected OGC-API schemas.

Issue #749, opened 2021-02-17
Implementing OGC API for processes


cehbrecht:
We want to implement the client part for the upcoming Rest API for Web Processing Services:
[OGC API - processes](https://ogcapi.ogc.org/processes/overview.html)

We can use this issue to collect ideas and resources.

huard:
Is there value in using the SDK automatically generated by https://app.swaggerhub.com/apis/geoprocessing/WPS/1.0-draft.5 ?

Export > Client SDK > Python

A lot of the boilerplate stuff seems to be handled already, but it's not clear how this meshes with the existing client implementation for features, collections, etc.

cehbrecht:
Spec:
https://ogcapi.ogc.org/processes/

Reference Implementations:
https://github.com/opengeospatial/ogcapi-processes/blob/master/implementations.adoc

Could start with [pygeoapi](https://docs.pygeoapi.io/en/latest/) as a test service.

Install locally:
https://docs.pygeoapi.io/en/latest/installation.html

Example Execute Request:
```
http://localhost:5000/processes/hello-world/jobs
```

Post request with payload:
```
{
"inputs": [
{"id": "name", "value": "hello"}
]
}
```

cehbrecht:
Collecting test data for offline tests:
* [52north](http://geoprocessing.demo.52north.org:8080/javaps/rest/): #748

cehbrecht:
> s there value in using the SDK automatically generated by https://app.swaggerhub.com/apis/geoprocessing/WPS/1.0-draft.5 ?

maybe as source of inspiration? Could copy+paste code?

fmigneault:
Do we want to add `getJson` methods and other 'reader' classes within the existing  `WebProcessingService` class?
I find many of the items should be handled by swapping the reader. (eg: when `getcapabilities` is called, instead the ogc-api reader would do `GET /processes` and parse JSON).

fmigneault:
For testing responses, I can open access to this one for the duration of the code-sprint: https://ogc-ades.crim.ca/ADES
The core OGC-API processes routes should be very close to the reference API with real processes.

huard:
@fmigneault Could you create static response files similar to https://github.com/geopython/OWSLib/pull/748 ?

Even better would be a function to automatically fetch static response files from the server, so that when the server gets an update, we can refresh the static test response files in the OWSLib repo.

cehbrecht:
Links from chat:
[...]

PR #750, opened 2021-02-17
initial ogcapi-processes implementation

cehbrecht:
This PR is first implementation for ogcapi-processes (#749).


cehbrecht:
Open issues:
* using/adding offline tests
* job status and cancel not tested
* handing POST and DELETE requests ... should go into base classes
* exception handling ... probably also part of base classes
* naming conventions `process_status` ...
* json inputs and outputs look simple ... but we might like to have something with python: https://pydantic-docs.helpmanual.io/datamodel_code_generator/
* server implementations are different ... not complete ... different responses ... POST blocked?

huard:
FYI: https://github.com/opengeospatial/ogcapi-processes/pull/121/files

owslib/ogcapi/processes.py
```diff
@@ -83,15 +100,28 @@ def execute(self, process_id: str, json: dict) -> dict:
         """
         # validate input
         m.Execute.parse_obj(json)
-
         path = f'processes/{process_id}/jobs'
         resp = self._request_post(path, json)
+
+        # Carsten, is it a server non-conformity that we can't just return m.Result.parse_obj(resp.json()) ?
```
cehbrecht, line 106: I was unsure were the `location` parameter is. On weaver it is returned via the header ... but it should also be the json.

tomkralidis:
Given there is an OAProc implementation now in master, should we close this?

tomkralidis:
@cehbrecht can we remove this branch?

PR #751, opened 2021-02-18
add implementation of wps-rest-client with ogc-api interface (#749)

fmigneault:
@cehbrecht 
I made a basic equivalent of the example wps-client for OGC-API.

### DIFF
--- a/examples/wps-rest-client.py
+++ b/examples/wps-rest-client.py
@@ -1,0 +1,251 @@
+#!/usr/bin/env python
+# -*- coding: ISO-8859-15 -*-
+# =============================================================================
+#
+# Authors : Francis Charette-Migneault (@fmigneault)
+#
+# =============================================================================
+
+import json
+import sys
+import getopt
+import os
+import yaml
+
+from owslib.ogcapi.processes import Processes
+
+
+def usage():
+    print("""
+    
+Usage: %s [parameters]
+
+Common Parameters for all request types
+-------------------
+
+    -u, --url=[URL] the base URL of the WPS - required
+    -r, --request=[REQUEST] the request type (GetCapabilities, DescribeProcess, Execute) - required 
+    -v, --verbose set flag for verbose output - optional (defaults to False)    
+    -o, --output=[FORMAT] format of the response to provide - optional {json, yaml}
+
+Request Specific Parameters
+---------------------------
+
+    DescribeProcess
+        -i, --identifier=[ID] process identifier - required
+    Execute
+        -d, --data, --json JSON file containing pre-made request to be submitted - required
+
+Examples
+--------
+python wps-rest-client.py -u https://ogc-ades.crim.ca/ADES -r GetCapabilities
+
+python wps-rest-client.py -u https://ogc-ades.crim.ca/ADES -r Processes
+
+python wps-rest-client.py -u https://ogc-ades.crim.ca/ADES -r DescribeProcess -i las2tif
+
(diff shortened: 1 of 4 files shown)


owslib/ogcapi/processes.py
```diff
@@ -65,29 +68,31 @@ def job_list(self, process_id: str) -> dict:
...
         """
 
         path = f'processes/{process_id}/jobs'
-        return self._request_post(path, json)
+        resp = self._request_post(path, json)
+        data = resp.json()
+        return resp.headers.get("Location", data["location"])
 
-    def _request_post(self, path: str, json: dict) -> dict:
+    def _request_post(self, path: str, json: dict) -> requests.Response:
```
cehbrecht, line 86: I would expect a json response

owslib/ogcapi/processes.py
```diff
@@ -124,3 +129,30 @@ def result(self, process_id: str, job_id: str) -> dict:
 
         path = f'processes/{process_id}/jobs/{job_id}/results'
         return self._request(path)
+
+    def monitor_execution(self, process_id: str = None, job_id: str = None, location: str = None,
```
cehbrecht, line 133: nice :smile:

cehbrecht (approved):

cehbrecht:
@fmigneault let me know when to merge ... I will add you to the developer lists.

cehbrecht:
Have a look at the `tests/test_ogcapi_processes_*` tests. They need to be adapted.

cehbrecht:
@fmigneault You are free to merge now when ready. Let me know when I should do the next iteration. 

I'm looking at a different owslib topic today:

owslib/ogcapi/processes.py
fmigneault: In this case, we actually need the response object as the OGC-API specifies that the job Execution should return the Status endpoint via `Location` header. I would also prefer the JSON object, but technically we need more than the body here.

examples/wps-rest-client.py
```diff
@@ -0,0 +1,255 @@
...
+    DescribeProcess, Execute, GetStatus
+        -i, --identifier=[ID] process identifier - required
+    Execute
+        -d, --data, --json JSON file containing pre-made request to be submitted - required
+    GetStatus
+        -j, --job JobID returned by the Execute request - required
+
+Examples
+--------
+python wps-rest-client.py -u https://ogc-ades.crim.ca/ADES -r GetCapabilities
```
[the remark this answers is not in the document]
fmigneault: We can add more examples here. That would be great in fact. `wps-client.py` also present many different example endpoints.
That endpoint is [crim-ca/weaver](https://github.com/crim-ca/weaver) but supports basic OGC-API Processes operations.

fmigneault:
@cehbrecht 

> 
> let me know when to merge ... I will add you to the developer lists.

I updated the test to work with a few processes that are always available in Weaver. 

Good to merge whenever desired.

owslib/ogcapi/processes.py
cehbrecht: We can make our own dict which contains all relevant information. But it would be handled in `owslib`.

cehbrecht:
merged ... I take a look at it tomorrow. I will also integrate #753.

fmigneault:
> merged ... I take a look at it tomorrow. I will also integrate #753.

#753 shouldn't affect since `api` and `conformance` methods are the only ones used by `processes` and they are left untouched.

PR #755, opened 2021-02-19
Data models for OGC API Processes

huard:
pydantic data models for the different objects transacted by the API.


cehbrecht (approved):

cehbrecht:
Tests are failing due to missing pydantic:
```
ModuleNotFoundError: No module named 'pydantic'
```

cehbrecht:
@huard @tomkralidis we probably want a seperation of these models? 

* core
  * collections
    * records
    * features
  * processes

cehbrecht:
Looks good :smile: I can update the `ogcapi.processes` to use this model.

huard:
Are models shared across collections, features, processes ? If not that separation sounds logical.

cehbrecht:
Looks `ForwardRef` is not support on py36 ... can we remove it? 
```
___________________ ERROR collecting owslib/ogcapi/models.py ___________________
  File "/home/runner/work/OWSLib/OWSLib/owslib/ogcapi/models.py", line 2, in <module>
    from typing import Any, List, Union, Optional, Dict, ForwardRef
ImportError: cannot import name 'ForwardRef'
```

huard:
Yes, we can remove it for now.

fmigneault:
@huard @cehbrecht `List["myref"]` is equivalent to `List[ForwardRef["myref"]]`
`ForwardRef` is for internal use

PR #758, opened 2021-02-19
using pydantic models in ogcapi.processes

cehbrecht:
this PR updates `ogcapi.processes` to use pydantic models (from #755):

* models are incomplete
* got it working with pygeoapi demo
* removed 52n demo


cehbrecht:
models look useful for validation ... but currently difficult ... each demo service response different.

Providing outputs with models looks fine but very similar to json.

owslib/ogcapi/models.py
```diff
@@ -129,7 +130,7 @@ class LiteralDataType(BaseModel):
 
 
 class ComplexDataType(BaseModel):
-    formats: List[FormatDescription]
+    formats: Optional[List[FormatDescription]]
```
huard, line 133: Do you need to add `= None` ?

owslib/ogcapi/models.py
```diff
@@ -271,9 +272,20 @@ class Subscriber(BaseModel):
...
+
+class LiteralOutput(BaseModel):
+    id: str
+    value: str
+
+
+class Result(BaseModel):
+    jobID: Optional[str]
+    location: Optional[str] = None
+    outputs: Optional[List[LiteralOutput]] = None
```
huard, line 291: This does not match the schema.

huard: I think what is missing is JobList

huard (approved):

owslib/ogcapi/models.py
cehbrecht: done

owslib/ogcapi/models.py
cehbrecht: done.

PR #763, opened 2021-03-02
Include custom headers in WMS requests

amercader:
I recently needed to interact with a WMS service which require authentication via a `User-agent` header (from [France's IGN](https://geoservices.ign.fr/documentation/donnees-ressources-wms-geoportail.html)).

When using `WebMapService` you can actually provide custom headers but they are not used later on when doing the actual HTTP requests:

    wms = WebMapService(
        'http://example.com/wms',
        version='1.3.0',
        headers={'User-agent': 'my-app/1.0'}
    )

    wms.getmap(...)  # Fails as not authorized

The `openURL()` helper already supports sending headers if provided, so it was just a matter of adding the WMS client instance headers (if present) to this function call.

Includes a small test.

### DIFF
--- a/owslib/map/wms111.py
+++ b/owslib/map/wms111.py
@@ -259,7 +259,7 @@ class WebMapService_1_1_1(object):
 
         self.request = bind_url(base_url) + data
 
-        u = openURL(base_url, data, method, timeout=timeout or self.timeout, auth=self.auth)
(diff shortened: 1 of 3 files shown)


amercader:
I get the same failure (`tests/test_ogcapi_features_pygeoapi.py::test_ogcapi_features_pygeoapi FAILED`) with latest master so not sure if it's related

PR #765, opened 2021-03-15
Extending MD_Keywords for where the keyword is a gmx:Anchor

adamml:
The Thesaurus had already been updated for this, but I have now added a case for keywords coming from a gmx:Anchor.

To facilitate this:

- I added a keyword attribute to MD_Keywords
- The MD_Keywords.keyword attribute is a list, and is populated by instances of a new class, keyword
- keyword has two attributes, name and URL 
- keyword is created for both cases of gmd:keyword - where the content is a gco:CharacterString and the url is therefore None, and for when the content is a gmx:Anchor and the url is defined by the href

### DIFF
--- a/owslib/iso.py
+++ b/owslib/iso.py
@@ -301,7 +301,7 @@ class CI_ResponsibleParty(object):
             self.role = _testCodeListValue(md.find(util.nspath_eval('gmd:role/gmd:CI_RoleCode', namespaces)))
 
 
-class keyword(object):
+class Keyword(object):
     """
     Class for complex keywords, with labels and URLs
     """
@@ -321,7 +321,9 @@ class MD_Keywords(object):
     def __init__(self, md=None):
     
         warnings.warn(
-            'The .keywords_object attribute will become .keywords proper in the next release',
+            'The .keywords_object attribute will become .keywords proper in the next release. '
+            '.keywords_object is a list of ibstances of the Keyword class. '
+            'See for https://github.com/geopython/OWSLib/pull/765 more details.',
             FutureWarning)
     
         if md is None:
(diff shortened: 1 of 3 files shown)


owslib/iso.py
```diff
@@ -298,22 +302,39 @@ def __init__(self, md=None):
             self.role = _testCodeListValue(md.find(util.nspath_eval('gmd:role/gmd:CI_RoleCode', namespaces)))
 
 
+class keyword(object):
```
huard, line 305: Use CamelCase for class names.

tomkralidis:
@adamml some comments:
- there are many parts of the commit with additions of blank lines and changes to `#noqa` markers.  Was this intended?
- this is a valuable enhancement given the functionality of anchors in ISO keywords.  We should really be planning to change the nature of `owslib.ogcapi.iso.MD_Keywords.keywords` from a list of strings to a list of objects. I would suggest changing the name `keyword` to `keywords_objects` and add a warning when initializing the `MD_Keywords` class that `keywords_objects` will become `keywords` proper in the next release

adamml:
@huard @tomkralidis Thanks for the review and the comments. I'll try and address them and comment back here

owslib/iso.py
adamml: Camel Case added in 1d5df8a, I will look to add a test now too

adamml: Tests added in 72c4148

adamml:
@huard, @tomkralidis Thanks again for reviewing the PR and making suggestions. I think I've addressed them all in a number of commits to my fork today. Let me know what you think and if there are any more steps to take before this PR can be merged.

Issue #768, opened 2021-04-16
Cartopy wmts KeyError issue : Content metadata for layer "AMSUA_NOAA15_..." already exists


CTOH-LEGOS:
#Since a while, I am using the wmts function of cartopy but since last week this function does not works returning this error:

KeyError: 'Content metadata for layer "AMSUA_NOAA15_Brightness_Temp_Channel_10" already exists'

The run code is coming from the cartopy web doc.

below the code:

`
 import matplotlib.pyplot as plt
 import cartopy.crs as ccrs

 def main():
     url = 'https://map1c.vis.earthdata.nasa.gov/wmts-geo/wmts.cgi'
     layer = 'VIIRS_CityLights_2012'

     fig = plt.figure()
     ax = fig.add_subplot(1, 1, 1, projection=ccrs.PlateCarree())
     ax.add_wmts(url, layer)
     ax.set_extent([-15, 25, 35, 60], crs=ccrs.PlateCarree())

     ax.set_title('Suomi NPP Earth at night April/October 2012')
     plt.show()

 if __name__ == '__main__':
     main()
`

The issue come from _buildMetadata method into  WebMapTileService class. The function called gather_layers raise an KeyError if there are duplicated element... But It should be better to add elements if they are not already in the dictionary called "contents"? And just add a warning for duplicate layers.

Below example to fix this issue:
`
 def gather_layers(parent_elem, parent_metadata):
     for index, elem in enumerate(parent_elem.findall(_LAYER_TAG)):
         cm = ContentMetadata( elem, parent=parent_metadata, index=index + 1,parse_remote_metadata=parse_remote_metadata)
         if cm.id:
             if not cm.id in self.contents:
                 self.contents[cm.id] = cm
             else:
                 warnings.war('Content metadata for layer "%s" '
                               'already exists' % cm.id)
         gather_layers(elem, cm)
`

Thank for help.
Fabien Blarel

nickdodds:
Hi, 

Also experiencing the exact same problem as above. I am trying to implement your solution, and was wondering if you had got this working. I did this by doing (you will notice your suggested change to the gather_layer function inside):

'''
from owslib.wmts import WebMapTileService

def custom_buildMetadata(self, parse_remote_metadata=False):
    ''' set up capabilities metadata objects '''

    self.updateSequence = self._capabilities.attrib.get('updateSequence')

    # serviceIdentification metadata
[...]

PR #769, opened 2021-04-16
changed all 'already exists' errors to warnings

speleophysics:
I was running into problems with using the NASA wmts service and came across these errors. Similar errors were pointed out in #619, which was fixed by pull request #656. However, I find that all of the "already exists" error messages in wmts.py were potentially problematic for the new configuration of the NASA server. Here I have converted all "already exists" errors into warnings, analogous to the change made in #656  for one of these error messages.

### DIFF
--- a/owslib/wmts.py
+++ b/owslib/wmts.py
@@ -233,8 +233,9 @@ class WebMapTileService(object):
                     parse_remote_metadata=parse_remote_metadata)
                 if cm.id:
                     if cm.id in self.contents:
-                        raise KeyError('Content metadata for layer "%s" '
+                        msg=('Content metadata for layer "%s" '
                                        'already exists' % cm.id)
+                        warnings.warn(msg, RuntimeWarning)
                     self.contents[cm.id] = cm
                 gather_layers(elem, cm)
         gather_layers(caps, None)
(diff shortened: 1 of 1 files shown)


greglucas:
Looks like this would close #768.

erin-raif:
I'm also having the same issue and this would fix it.

lgolston:
Yes, I'd like to echo this issue. Simply running:
```
from owslib.wmts import WebMapTileService
wmts = WebMapTileService("http://map1c.vis.earthdata.nasa.gov/wmts-geo/wmts.cgi")
```
Gives
`KeyError: 'Content metadata for layer "MODIS_Combined_MAIAC_L2G_AerosolOpticalDepth" already exists'`

We are seeing this over at cartopy, where the WMTS interface is currently broken as a result.

greglucas:
This would be nice to get in to avoid these recurring errors that happen when NASA updates its server content. This does seem like an issue with the NASA servers and not here, but this would at least allow programs to keep running and just warn about bad data rather than erroring on bad data.

I sent an e-mail to NASA support about this issue as well to try and make them aware of it on their end.

greglucas:
Ping again to try and get this looked at.

rcomer:
It would be great if this PR could get a maintainer decision, rather than being closed by the bot.

geographika:
This looks fine to me to merge. Several other modules use warnings rather than errors. 
Unless any objections @tomkralidis I can merge this in the next few days.

tomkralidis (approved):

geographika:
Thanks @speleophysics - merging.

PR #770, opened 2021-04-22
Make most of Authentication's attributes private

### LINKED ISSUE #762: AttributeError: 'Authentication' object has no attribute '_auth_delegate'
Hi,

while authenticating towards a CSW Server using OWSLib 0.23.0 I get the following error:

```
  File "/foo/bar/pipeline/server.py", line 57, in <module>
    pipeline = Pipeline()
  File "/foo/bar/pipeline/pipeline.py", line 41, in __init__
    self.cswConnector.connect()
  File "/foo/bar/pipeline/csw/CSWConnector.py", line 16, in connect
    self.csw = CatalogueServiceWeb(url, username=username, password=password)
  File "/foo/bar/venv/lib/python3.9/site-packages/owslib/csw.py", line 73, in __init__
    self.auth = auth or Authentication(username, password)
  File "/foo/bar/venv/lib/python3.9/site-packages/owslib/util.py", line 875, in __init__
    self.username = username
  File "/foo/bar/venv/lib/python3.9/site-packages/owslib/util.py", line 894, in username
    if self.auth_delegate is not None:
  File "/foo/bar/venv/lib/python3.9/site-packages/owslib/util.py", line 983, in auth_delegate
    return self._auth_delegate
AttributeError: 'Authentication' object has no attribute '_auth_delegate'
```

As far as I can tell, the issue is that the `Authentication` class accesses members that are not yet initialized.
I could track down the following commit that introduced the respective code:
https://github.com/geopython/OWSLib/commit/00b07e7d3b57105677404d250a2763a6550e7d86

The error can be resolved by initializing the internal members with None at the beginning of the `__init__` method with:
```
def __init__(self, username=None, password=None,
                 cert=None, verify=True, shared=False,
                 auth_delegate=None):

        # ------------- fix/workaround
        self._username = None
        self._password = None
        self._cert = None
        self._verify = None
        self._auth_delegate = None
        # ------------- fix end
       
        self.shared = shared
        self.username = username
        self.password = password
        self.cert = cert
        self.verify = verify
        self.auth_delegate = auth_delegate
```

Best,
Adrian

maximlt:
Except for `shared` which is not a property.

Fixes #762

### DIFF
--- a/owslib/util.py
+++ b/owslib/util.py
@@ -872,11 +872,11 @@ class Authentication(object):
             username/password arguments.
         '''
         self.shared = shared
-        self.username = username
-        self.password = password
-        self.cert = cert
-        self.verify = verify
-        self.auth_delegate = auth_delegate
+        self._username = username
+        self._password = password
+        self._cert = cert
+        self._verify = verify
+        self._auth_delegate = auth_delegate
 
     @property
     def username(self):
--- a/owslib/util.py
+++ b/owslib/util.py
@@ -878,6 +878,14 @@ class Authentication(object):
         self._verify = verify
         self._auth_delegate = auth_delegate
 
+        # Trigger the setters to validate the parameters. This couldn't be done directly
+        # since some parameters are mutually exclusive.
+        self.username = username
+        self.password = password
+        self.cert = cert
+        self.verify = verify
+        self.auth_delegate = auth_delegate
+
     @property
     def username(self):
         if self.shared:


cehbrecht:
@maximlt I find the code of `Authentication` a bit confusing. I think the current fix in this PR does not solve the issue. The proposed fix in #762 looks better (init `self._username = None` etc)

cehbrecht (changes requested):
I think the proposed fixed in #762 looks better.

maximlt:
@cehbrecht 

Thanks for having a look at this one.

If I'm correct, the fixed suggested in #762 is:
```python
def __init__(self, username=None, password=None,
                 cert=None, verify=True, shared=False,
                 auth_delegate=None):

        # ------------- fix/workaround
        self._username = None
        self._password = None
        self._cert = None
        self._verify = None
        self._auth_delegate = None
        # ------------- fix end
       
        self.shared = shared
        self.username = username
        self.password = password
        self.cert = cert
        self.verify = verify
        self.auth_delegate = auth_delegate
```

Which is pretty much what I have done, except that in #762 each attribute is assigned None (e.g. `self._username = None`). This is wrong, because, for instance  `Authentication(username="login", password="secret")` would lead to both `username` and `password` set to None.

 As for the overall code of `Authentication` being confusing, I tend to agree with you but I don't have the bandwith currently to work on refactoring it, fixing it was my priority :)

cehbrecht:
@maximlt I think the original code wants to do the initialization via the `property` method, so we still need something like this:
```
self.shared = shared
self.username = username
```

tomkralidis (approved):

maximlt:
@cehbrecht indeed, now the setters are not triggered during the initialization, which is unfortunate. I'll make a very simple PR to adjust that, and let you both decide if you're fine with it !

cehbrecht:
test failure not related to this PR.

cehbrecht (approved):
Looks good to me.

cehbrecht:
ping @eric-spitler

eric-spitler:
Thanks for the ping, No concerns from me.

PR #779, opened 2021-06-07
OAFeat: cast items bbox parameter values to string

tomkralidis:


tomkralidis:
Test failures are unrelated.

Issue #128, opened 2014-03-03
Implement remaining spatial filters in owslib.fes


kwilcox:
Only BBOX is currently implemented

lpinner:
+1 for this to be implemented.

I'm currently working around with some simple envelope only code based on the owslib.fes.BBox code:

```python
## owslib.fes only supports BBox spatial filter, so add a few more
## Modified by Luke Pinner (2014) from original owslib.fes.BBox code
class SpatialOp(OgcExpression):
    """Construct a BBox, two pairs of coordinates (west-south and east-north)"""
    def __init__(self, operator, bbox, crs=None):
        self.operator = operator
        self.bbox = bbox
        self.crs = crs
    def toXML(self):
        tmp = etree.Element(util.nspath_eval(self.operator, namespaces))
        etree.SubElement(tmp, util.nspath_eval('ogc:PropertyName', namespaces)).text = 'ows:BoundingBox'
        tmp2 = etree.SubElement(tmp, util.nspath_eval('gml:Envelope', namespaces))
        if self.crs is not None:
            tmp2.set('srsName', self.crs)
        etree.SubElement(tmp2, util.nspath_eval('gml:lowerCorner', namespaces)).text = '%s %s' % (self.bbox[0], self.bbox[1])
        etree.SubElement(tmp2, util.nspath_eval('gml:upperCorner', namespaces)).text = '%s %s' % (self.bbox[2], self.bbox[3])
        return tmp

class Contains(SpatialOp):
    def __init__(self, bbox, crs=None):
        SpatialOp.__init__(self,'ogc:Contains',bbox,crs)

class Disjoint(SpatialOp):
    def __init__(self, bbox, crs=None):
        SpatialOp.__init__(self,'ogc:Disjoint',bbox,crs)

class Within(SpatialOp):
    def __init__(self, bbox, crs=None):
        SpatialOp.__init__(self,'ogc:Within',bbox,crs)
```

Zeitsperre:
Is this issue still a work in progress ?

huard:
Fixed for WFS2.0 and POST requests in #780

PR #780, opened 2021-06-14
Add support for topological filters

huard:
This PR allows users to filter WFS requests using spatial filters: Intercepts, Contains, Within, Touches, Disjoint, Overlaps, Equals. 

- Add spatial filters in fes2.py;
- Create gml:Point object;

# Notes

- Only works with WFS 2.0 and POST. 
- The `set_filter` method in GetFeature is weird. It expects a GetFeature xml request, instead of just the Filter node.  I've added a shortcut to allow passing a filter XML object directly. Suggestions welcome. I suggest that when passing `filter` to getfeature, we only pass the filter node, not the getfeature node. 
- I've tested it on our own geoserver instance, but unclear how to test this actually works without a live instance somewhere. 
- The implementation bypasses the FilterRequest class, which seemed to target CSW.
- I used dataclasses, but that's only available from 3.7 onward so added it to dependencies if python < 3.7.

# Related
#128

### DIFF
--- a/owslib/feature/postrequest.py
+++ b/owslib/feature/postrequest.py
@@ -173,8 +173,12 @@ class PostRequest_2_0_0(PostRequest):
 
         Cannot be used with set_bbox() or set_featureid().
         """
-        f = etree.fromstring(filter)
-        sub_elem = f.find(util.nspath("Filter", FES_NAMESPACE))
+        if isinstance(filter, str):
+            f = etree.fromstring(filter)
+            sub_elem = f.find(util.nspath("Filter", FES_NAMESPACE))
+        else:
+            sub_elem = filter
+
         self._query.append(sub_elem)
 
     def set_maxfeatures(self, maxfeatures):
--- a/owslib/fes2.py
+++ b/owslib/fes2.py
@@ -18,6 +18,7 @@ Supports version 2.0.2 (09-026r2).
 from owslib.etree import etree
 from owslib import util
 from owslib.namespaces import Namespaces
+from abc import ABCMeta, abstractmethod
 
 
 # default variables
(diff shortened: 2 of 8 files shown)


huard:
Example usage:
```python
from owslib.fes2 import Intersects
from owslib.gml import Point
from owslib.util import element_to_string
import geojson


def test_intersects():
    p = Point(id="qc", srsName="http://www.opengis.net/gml/srs/epsg.xml#4326", pos=[-71, 46])
    f = Intersects(propertyname="the_geom", geometry=p)
    intersects = f.toXML()
    wfs = WebFeatureService("https://pavics.ouranos.ca/geoserver/wfs", version="2.0.0")
    resp = wfs.getfeature(typename="public:canada_admin_boundaries", outputFormat="application/json",
                   propertyname="PRENAME", method="POST", filter=intersects)
    out = geojson.load(resp)
    assert out["totalFeatures"] == 1
    assert out["features"][0]["properties"]["PRENAME"] == "Quebec"
```

tomkralidis:
Thanks @huard.  There are no plans to drop 3.6 anytime soon given it is the default Python version on major OSs (i.e. bionic).

huard:
Ok, will modify the PR.

huard:
As mentioned, at the moment the `<Filter>` tag is built into each topological filter. I think ideally, users should be able to compose filters with AND, OR operations. This would require a smarter composition mechanism.

huard:
@kwilcox @lpinner @jorisvandenbossche Is this what you were looking for in issue #128 ?

huard:
@tomkralidis Ready for review.

tomkralidis (approved):

tomkralidis:
Thanks @huard !

GabrielFerrariR:
is there any provision for stable implementation of geometry types other than Point?

huard:
Not from my part. This PR was meant to solve a problem I had.

PR #784, opened 2021-07-05
iterative download of outputs

jachym:
This pull request:
* introduces `iter_content` method for the `ResponseWrapper` object - this will make reading of large outputs more safe
* outputs will not be returned back as whole data object downloaded in one chunk, but rather the objects, which can be downloaded in more safe way

### DIFF
--- a/owslib/wps.py
+++ b/owslib/wps.py
@@ -116,6 +116,7 @@ from xml.dom.minidom import parseString
 from owslib.namespaces import Namespaces
 from urllib.parse import urlparse
 import warnings
+import math
 
 # namespace definition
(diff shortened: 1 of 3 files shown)


owslib/wps.py
```diff
@@ -1438,7 +1449,7 @@ def retrieveData(self, username=None, password=None, headers=None, verify=True,
         # Useful when running local tests during development.
         if url.startswith("file://"):
             with open(url[7:]) as f:
-                return f.read()
```
huard, line 1441: I suspect this is what causes the remaining test failure.

PR #789, opened 2021-07-29
Update pyproj syntax and fix some issue in WFS

cheginit:
This PR includes the following improvements/fixes:
1. Update `pyproj` syntax for the transform operation according to the proposed style on their [website](https://pyproj4.github.io/pyproj/stable/gotchas.html) to fix the deprecation warning.
2. Fix an issue in `WFS` where a list addition operation was not assigned to the intended variable.
3. Remove the unnecessary `else` in an if-block, since an exception is being raised.

### DIFF
--- a/owslib/feature/wfs100.py
+++ b/owslib/feature/wfs100.py
@@ -6,6 +6,7 @@
 # $Id: wfs.py 503 2006-02-01 17:09:12Z dokai $
 # =============================================================================
 
+import itertools
 from owslib import util
 
 from io import BytesIO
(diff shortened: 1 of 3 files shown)


huard (approved):
Could you add a test that covers the bug this fixes to avoid future regression?

cheginit:
Sure, I'll work on it.

cheginit:
I am having difficulty finding a WFS service with `Operations` for `FeatureType`s themselves not just the parent `FeatureTypeList`. I tried a couple of services but their `elem.findall(nspath("Operations/*")` are empty. Do you know such WFS service?

cheginit:
Sure. Will do that.

cheginit:
Is there anything else that you would like to include in this PR?

huard:
Ideally all tests would pass before merging, but I assume the failure I see is unrelated to your PR?

@tomkralidis Any idea?

tomkralidis:
+1 to merge this PR given the test failures are unrelated.  If the unrelated test failures persist we should address as part of another fix to the tests.

cheginit:
Yes, it seems that the tests that are failing require another PR. I confirmed that the tests that I added for this PR pass without any issue.

cheginit:
@huard I am wondering what's your plan for this PR?

PR #796, opened 2021-09-08
Add maximumMegabytes.

fmigneault:
Transfer `maximumMegabytes` attribute if available in WPS ComplexData definition. 
This is useful for applications that want to make use of objects generated by parsing a `DescribeProcess` response, but is not explicitly employed by the code.

### DIFF
--- a/owslib/wps.py
+++ b/owslib/wps.py
@@ -1062,7 +1062,7 @@ class ComplexData(object):
         self.mimeType = mimeType
         self.encoding = encoding
         self.schema = schema
-        self.maximumMegabytes = maximumMegaBytes
+        self.maximumMegabytes = maximumMegaBytes  # defined only if provided by ProcessDescription ComplexData
 
 
 class InputOutput(object):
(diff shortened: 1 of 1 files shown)


huard (approved):
The default is None, does that mean that by default there is no maximum size ?
Could you add a line in a docstring to explain what this does ?
Is this something we should test with emu ?

fmigneault:
@huard 

> The default is None, does that mean that by default there is no maximum size ?

Yes, this is the default behaviour. 
In the XML, `<ComplexData>` can be used instead of `<ComplexData maximumMegabytes="<max_mb>">`. 
Note that I only make the parameter available. 
Regardless of it being present or not, there is no special operation or validation during the Execute request in OWSLib at the moment (I believe it should probably be done on the pywps side anyway). 

> Could you add a line in a docstring to explain what this does ?

Certainly.

> Is this something we should test with emu ?

That would be a good idea. 
I can see that `pywps` checks for a server-wide max-size rather than per-process here: 
https://github.com/geopython/pywps/blob/ec6f6b/pywps/inout/basic.py#L539-L563
This value is the same one employed to generate the template: 
https://github.com/geopython/pywps/blob/711219/pywps/templates/1.0.0/describe/main.xml#L30-L32
So should be working as expected.

PR #794, opened 2021-08-27
fix WPS processVersion attribute retrieval during DescribeProcess request

fmigneault:
When a `GetCapabilities` response is obtained, attribute `wps:processVersion` of all process offerings are nested under `wps:Process` elements. This makes parsing of the "root element" as namespace `wps` for each process. 

```xml
<wps:Capabilities service="WPS"
                  version="1.0.0"
                  xml:lang="en-US"
                  xsi:schemaLocation="http://www.opengis.net/wps/1.0.0 ../wpsGetCapabilities_response.xsd"
                  updateSequence="1">
    [...]
	<wps:ProcessOfferings>
		<wps:Process wps:processVersion="0.1">
			<ows:Identifier>raven</ows:Identifier>
			<ows:Title>
		</wps:Process>
 		[...]
	</wps:ProcessOfferings>	
	[...]
</wps:Capabilities>
```

On the other hand, the `DescribeProcess` response provides attribute `wps:processVersion` under element  `ProcessDescription` of the retrieved process(es) (without `wps` namespace), making the "root element" namespace an empty string. 
```xml
<wps:ProcessDescriptions xsi:schemaLocation="http://www.opengis.net/wps/1.0.0 ../wpsDescribeProcess_response.xsd"
                         service="WPS"
                         version="1.0.0"
                         xml:lang="en-US">
	<ProcessDescription wps:processVersion="0.1"
	                    storeSupported="true"
	                    statusSupported="true">
		<ows:Identifier>raven</ows:Identifier>
 		[...]
	</ProcessDescription>
</wps:ProcessDescriptions>
```

The `Process` class was instantiated by always picking the active namespace from the root element, whether it was `wps` or `""`, but `wps:processVersion` always has the `wps` namespace. 

This PR resolves retrieval of `processVersion` attribute following a `DescribeProcess` request. 

Ex before: 
```python 
wps = WebProcessingService("<url>")
print([p.processVersion for p in wps.processes if p.identifier == "raven"])
# ['0.1']   # works because of 'wps:Process' 

process = wps.describeprocess("raven")
print(process.processVersion)
# None   # fails because of 'ProcessDescription' instead of 'wps:ProcessDescription'
```

With PR:
```python
print(process.processVersion)
# "0.1"
```

PR #797, opened 2021-10-15
Fix tests

cehbrecht:
This PR fixes the test suite.

Changes:
* relaxed conditions for ogcapi features test
* `xfail` broken tests ... wfs service returns exception
* fixed wfs100 `boundingBoxWGS84` 
* pep8 for `iso.py`
* fixed wps test: use `get_namespace()` instead of `elem.nsmap` (worked only with `lxml`). See PR #794.

### DIFF
--- a/owslib/feature/wfs100.py
+++ b/owslib/feature/wfs100.py
@@ -395,9 +395,9 @@ class ContentMetadata(AbstractContentMetadata):
         self.boundingBoxWGS84 = None
 
         if b is not None and srs is not None:
-            wgs84 = pyproj.Proj("epsg:4326")
+            wgs84 = pyproj.CRS.from_epsg(4326)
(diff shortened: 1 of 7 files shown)


tomkralidis:
> @tomkralidis I have questions about our test suite:
> 
>     * Can we make `lxml` a mandatory requirement?
> 

We are supposed to be able to cover both lxml and non-lxml installations.  Would we be missing some test cases for those not using lxml in their OWSLib install?

>     * Should we mark "online" tests with `pytest xfail`? We can in addition use `assert service_ok is True`.
>       
>       * https://docs.pytest.org/en/6.2.x/skipping.html
> 
>     * we could run in addition "online" tests separately in our CI. Online tests would be optional and allowed to fail.

+1

PR #799, opened 2021-11-02
adding check for gmx:Anchor

epifanio:
* make sure `accessconstraints`, `useconstraints` and `identifier`  values are picked when available as `gmx:Anchor` instead of `gco:CharacterString`
* extend `resourcelanguage`,  MI_Instrument

Some background info:

https://www.ngdc.noaa.gov/wiki/index.php/Character_String_Extended_Types -  """ With the gmx namespace added, gmx:Anchor is a valid substitute for any gco:CharacterString in the ISO XML"""


huard (commented):
I'm not familiar with this part of OWSLib, but is there a way to add meaningful units tests for this ?

pvgenuchten:
suggestion it to add a function for extracting characterstring|anchor, with doc and xpath to parent element as input

unit test should be able to extract (link and) string or nill from 

either

```
<xxx gco:nilReason="missing">
            <gco:CharacterString/></xxx>
```
or
```
<xxx>
 <gco:CharacterString>Nederlands metadata profiel op ISO 19119 voor services 2.1.0</gco:CharacterString>
</xxx>
```
or
```
<xxx>
  <gmx:Anchor xlink:href="http://www.opengis.net/def/crs/EPSG/0/28992">Rijks Driehoeksstelsel</gmx:Anchor>
</xxx>
```

tomkralidis:
+1/agree with @huard to add a couple of unit tests to test functionality between `gco:CharacterString` and `gmx:Anchor`.

What would be even better is a local utility function which can detect one or the other, and be reused across those ISO elements support either.

epifanio:
> 

Working on this issue this week. Thanks for the comments!

pvgenuchten:
Hi, it looks like a good start, thanx!

Interesting aspect of using an anchor is that it typically does not return a string, but a combination of xlink:href and string (or either one). 
the owslib model may not be ready to capture {label:example,link:foo}, alternative could be to store it as ```[ label ]( href )``` (markdown style)

PR #807, opened 2022-03-12
Revert to previous pyproj version

kalxas:


kalxas:
This now only has 3 fails due to remote WFS service error

tomkralidis:
CI failures unrelated.

musicinmybrain:
Is there an issue tracking the problems you’re having with pyproj 3.3, or a known path toward resolving them? Thanks!

sbrunner:
Why this limitation?

fmigneault:
@tomkralidis 
Would it be possible to have a `0.26.1` fix with the reverted requirement in https://github.com/geopython/OWSLib/commit/a6105226f1c8466c3b1e9998747d0dbf01434129 before all the *OGC API - Processes* items are added?

tomkralidis:
@fmigneault we should be able to cut a release during the FOSS4G code sprint.  Is there forthcoming work coming for OGC API - Processes?

fmigneault:
I don't know if there will be significant work about *OGC API - Processes*, but I think a patch release should be done beforehand such that `pyproj` can work once again on Python 3.10 without possibility introducing new issues from new features.

Issue #810, opened 2022-03-22
Need of additional url request parameters (e.g. for API keys)


danduk82:
Some servers require the client to add some additional URL request parameters (for example API keys).

I would like to provide a Pull Request that adds this feature for WMS and WFS services (and harmonize WMTS).

see my PR: https://github.com/geopython/OWSLib/pull/811

Use-case
-------------

**As:** a user
**I want:** use OWSLib as a client for [geoserver OGC server](http://geoserver.org/) with the plugin [authkey](https://docs.geoserver.org/stable/en/user/community/authkey/index.html) which requires the possibility to add `&authkey=<some-uuid>` in the request parameters
**Because:** the server that I must use uses this kind of API keys for access and currently I cannot use OWSLib

danduk82:
@tomkralidis may I ask if you have any opinion on this issue and the respective PR? Just to get a reaction from the maintainer side please.

jmckenna:
This seems like a useful addition to me.

tomkralidis:
In the past, OWSLib has typically delegated extra parameters to be part of the base URL in the constructor.  Example:

```python
from owslib.wms import WebMapService

w = WebMapService('https://example.org/wms?auth_key=123')
```

From a pure WMS standpoint, given the above, the WMS specification functions as expected without needing to manage extra parameters.  Does this approach work for your use case?

danduk82:
@tomkralidis  actually I didn't think that enough and discussing with a colleague also my PR doesn't take into account the fact that the ressource URLs might not lead to the same server.

So indeed, your KISS solution is better. I will close the PR. :+1:

PR #811, opened 2022-03-22
WMS,WFS: add custom request parameters

danduk82:
This pull request:
- adds the possibility to add additional URL request parameters for WMS and WFS services.
- moves the `vendor_kwargs` parameters of WMTS methods in the constructor

Fixes https://github.com/geopython/OWSLib/issues/810

breaking changes:
- the parameter `vendor_kwargs` in wmts functions is removed, and replaced with a `**kwargs` logic in the wmts main constructor. The reason for this is that some vendors require additional parameters already in the GetCapabilities request. Also,  if all the requests need to be "decorated" with custom parameters, it makes sense to put them already in the constructor.

### DIFF
--- a/owslib/feature/common.py
+++ b/owslib/feature/common.py
@@ -29,8 +29,7 @@ class WFSCapabilitiesReader(object):
         self._infoset = None
         self.vendor_kwargs = {}
         if kwargs:
-            for kw in kwargs:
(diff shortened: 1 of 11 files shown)


danduk82:
closing this PR as the additional parameters should be added to the URL in the constructor directly

PR #812, opened 2022-03-28
pass timeout to WCS constructors

tomkralidis:


tomkralidis:
FYI CI failures are unrelated.

PR #817, opened 2022-04-18
fix tests

tomkralidis:
This PR provides some test fixes and superfluous warnings.

### DIFF
--- a/owslib/owscontext/atom.py
+++ b/owslib/owscontext/atom.py
@@ -215,7 +215,7 @@ def parse_entry(entry_node):
         elif link.get("rel") == "via" and not is_empty(link):
             links_via.append(link)
         else:
-            log.warn(
+            log.warning(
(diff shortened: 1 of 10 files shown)


tomkralidis:
FYI there's one remaining failing test: https://github.com/geopython/OWSLib/runs/6069051764?check_suite_focus=true#step:6:328

I'm not sure what to make of this, other than a potentially misconfigured WFS2 instance (cc @huard)

cehbrecht (approved):

cehbrecht:
> I'm not sure what to make of this, other than a potentially misconfigured WFS2 instance (cc @huard)

If we can not fix it now, we could use `pytest.xfail`.

tomkralidis:
Thanks @cehbrecht.  Merging.  Latest CI failure is intermittent.

PR #818, opened 2022-04-20
fix pygeoapi OGC API tests

tomkralidis:
This PR updates the OGC API pygeoapi-based tests for OACov and OARec to working demo servers.

### DIFF
--- a/tests/test_ogcapi_coverages_pygeoapi.py
+++ b/tests/test_ogcapi_coverages_pygeoapi.py
@@ -4,7 +4,7 @@ import pytest
 
 from owslib.ogcapi.coverages import Coverages
 
-SERVICE_URL = 'https://dev.api.weather.gc.ca/coverages-demo'
+SERVICE_URL = 'https://demo.pygeoapi.io/master/'
 
 
 @pytest.mark.online
@@ -13,7 +13,7 @@ SERVICE_URL = 'https://dev.api.weather.gc.ca/coverages-demo'
 def test_ogcapi_coverages_pygeoapi():
     w = Coverages(SERVICE_URL)
 
-    assert w.url == 'https://dev.api.weather.gc.ca/coverages-demo/'
+    assert w.url == SERVICE_URL
     assert w.url_query_string is None
 
     api = w.api()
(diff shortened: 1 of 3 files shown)


tomkralidis:
FYI CI test failure is intermittent.

Aside: we have a number of warnings of type:

```python
 tests/test_wmts.py:94
  /home/runner/work/OWSLib/OWSLib/tests/test_wmts.py:94: PytestUnknownMarkWarning: Unknown pytest.mark.online - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    @pytest.mark.online
```

Examples: https://github.com/geopython/OWSLib/runs/6087949336?check_suite_focus=true#step:6:352

@cehbrecht any thoughts here on these warnings?

cehbrecht:
> @cehbrecht any thoughts here on these warnings?

We can use `pytest` markers to filter tests. `online` means that the test needs external resources. 
https://docs.pytest.org/en/latest/how-to/mark.html 

Somehow the config for the `pytest` markers is missing. We could add a `pytest.ini` as described in the `pytest` docs.

tests/test_ogcapi_coverages_pygeoapi.py
```diff
@@ -4,7 +4,7 @@
 
 from owslib.ogcapi.coverages import Coverages
 
-SERVICE_URL = 'https://dev.api.weather.gc.ca/coverages-demo'
+SERVICE_URL = 'https://demo.pygeoapi.io/master/'
```
cehbrecht, line 7: 👍

cehbrecht (approved):

tomkralidis:
> > @cehbrecht any thoughts here on these warnings?
> 
> We can use `pytest` markers to filter tests. `online` means that the test needs external resources. https://docs.pytest.org/en/latest/how-to/mark.html
> 
> Somehow the config for the `pytest` markers is missing. We could add a `pytest.ini` as described in the `pytest` docs.

Thanks @cehbrecht; I've added this to the PR in `tox.ini`, which removes the warnings.

PR #825, opened 2022-07-06
add support for OGC API - Processes

tomkralidis:


owslib/ogcapi/processes.py
```diff
@@ -0,0 +1,46 @@
...
+import logging
+
+from owslib.ogcapi import Collections
+from owslib.util import Authentication
+
+LOGGER = logging.getLogger(__name__)
+
+
+class Processes(Collections):
+    """Abstraction for OGC API - Features"""
```
kalxas, line 18: s/Features/Processes/g

tomkralidis: fixed

PR #861, opened 2023-02-13
Raise an error if GetCapabilities requests don't return XML

geographika:
Currently if a GetCapabilities request doesn't return an XML document it raises cryptic error messages such as:

```
  File "owslib\feature\wfs200.py", line 164, in _buildMetadata
    for elem in self._capabilities.find(nspath("OperationsMetadata"))[:]:
TypeError: 'NoneType' object is not subscriptable
```

This pull request checks to see if the server response `Content-Type` is XML, and if not it raises an exception that includes the URL and response. If the response is HTML it attempts to parse the body of the document to use as the error message. This is particularly useful when working with MapServer which returns errors in `text/html` format. E.g.

```
http:///example.org/?service=WFS&request=GetCapabilities&version=2.0.0 responded 
with Content-Type 'text/html': 'msCGILoadMap(): Web application error. CGI variable "map" is not set.'
```

Currently any `Content-Type` which has any match/partial match against the following list will be considered as XML: `
['text/xml', 'application/xml', 'application/vnd.ogc.wms_xml']`

It may be better to just check for `xml` in case there are types I've missed?

In addition the `strip_bom` was only previously used for WMS GetCapabilities requests - this change means it is applied to all service types (WFS, SOS, WCS, WMTS).

### DIFF
--- a/owslib/coverage/wcsBase.py
+++ b/owslib/coverage/wcsBase.py
@@ -11,7 +11,7 @@
 
 from urllib.parse import urlencode, parse_qsl
 from owslib.etree import etree
-from owslib.util import Authentication, openURL
+from owslib.util import Authentication, openURL, getXMLTree
 
 
 class ServiceException(Exception):
@@ -118,7 +118,7 @@ class WCSCapabilitiesReader(object):
         """
         request = self.capabilities_url(service_url)
         u = openURL(request, timeout=timeout, cookies=self.cookies, auth=self.auth, headers=self.headers)
-        return etree.fromstring(u.read())
+        return getXMLTree(u)
 
     def readString(self, st):
         """Parse a WCS capabilities document, returning an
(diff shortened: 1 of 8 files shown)


geographika:
Rebased against main - all tests passing.

geographika:
Failing CI is due to not being able to submit to coveralls. Seems unrelated to this change. 

```
 Submitting coverage to coveralls.io...
Error running coveralls: Could not submit coverage: 422 Client Error: Unprocessable Entity for url: https://coveralls.io/api/v1/jobs
  File "/opt/hostedtoolcache/Python/3.11.10/x64/lib/python3.11/site-packages/coveralls/api.py", line 298, in submit_report
Received 422 submitting job via Github Actions. By default, coveralls-python uses the "github" service name, which requires you to set the $GITHUB_TOKEN environment variable. If you want to use a COVERALLS_REPO_TOKEN instead, please manually override $COVERALLS_SERVICE_NAME to "github-actions". For more info, see https://coveralls-python.readthedocs.io/en/latest/usage/configuration.html#github-actions-support
    response.raise_for_status()
  File "/opt/hostedtoolcache/Python/3.11.10/x64/lib/python3.11/site-packages/requests/models.py", line 1024, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
```

PR #863, opened 2023-02-23
use only lxml for XML handling

tomkralidis:


tomkralidis:
CI failure is unrelated.

kalxas (approved):

PR #868, opened 2023-04-07
add OpenSearch client implementation

tomkralidis:
This PR adds support for [OpenSearch 1.1](https://github.com/dewitt/opensearch), which is in use among various EO service providers.  The handler supports JSON and Atom based responses (big kudos to @allixender for `owslib.owscontext.atom`, which enabled easy reuse here for Atom integration!).


kalxas (approved):

kalxas:
CI failures unrelated

PR #883, opened 2023-07-07
Fix pip install command in README

### LINKED ISSUE #882: Copy-Paste Error in README
In the README file it states, that for installation you should run:

https://github.com/geopython/OWSLib/blob/bc88ee1a6509746cff1cccfbad6e6501bae4c256/README.md?plain=1#L24

I guess the part in the README was just copied from the other lib and the author forgot to replace the pip install command.

JasperJuergensen:
Fixes the pip install command in README file.

Closes #882

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,11 +17,11 @@ numerous OGC Web Service interfaces.
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


tomkralidis:
Since fixed

PR #891, opened 2023-09-24
update CI (drop Python 3.7, add 3.10/3.11)

tomkralidis:


.github/workflows/main.yml
```diff
@@ -20,7 +20,6 @@ jobs:
     - name: Install packages
       run: |
         sudo apt-get -y install pandoc
-      if: matrix.python-version == 3.7
```
kalxas, line 23: This was placed here to save resources, maybe just move to 3.8?

kalxas (approved):

PR #899, opened 2023-12-16
add OGC API - EDR client

tomkralidis:
FYI this PR should be merged once the pygeoapi demo server is updated via [this PR](https://github.com/geopython/pygeoapi/pull/1431).

### DIFF
--- a/owslib/ogcapi/edr.py
+++ b/owslib/ogcapi/edr.py
@@ -1,0 +1,98 @@
+# =============================================================================
+# Copyright (c) 2023 Tom Kralidis
+#
+# Author: Tom Kralidis <tomkralidis@gmail.com>
+#
+# Contact email: tomkralidis@gmail.com
+# =============================================================================
+
+from io import BytesIO
+import logging
+from typing import BinaryIO
+
+from owslib.ogcapi.features import Features
+from owslib.util import Authentication
+
+LOGGER = logging.getLogger(__name__)
+
+
+class EnvironmentalDataRetrieval(Features):
+    """Abstraction for OGC API - Envirionmental Data Retrieval"""
+
+    def __init__(self, url: str, json_: str = None, timeout: int = 30,
+                 headers: dict = None, auth: Authentication = None):
+        __doc__ = Features.__doc__  # noqa
+        super().__init__(url, json_, timeout, headers, auth)
+
+    def data(self) -> list:
+        """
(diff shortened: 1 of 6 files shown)


owslib/ogcapi/edr.py
```diff
@@ -0,0 +1,98 @@
...
+from typing import BinaryIO
+
+from owslib.ogcapi.features import Features
+from owslib.util import Authentication
+
+LOGGER = logging.getLogger(__name__)
+
+
+class EnvironmentalDataRetrieval(Features):
+    """Abstraction for OGC API - Envirionmental Data Retrieval"""
```
huard, line 20:
```suggestion
    """Abstraction for OGC API - Environmental Data Retrieval"""
```

owslib/ogcapi/edr.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def __init__(self, url: str, json_: str = None, timeout: int = 30,
+                 headers: dict = None, auth: Authentication = None):
+        __doc__ = Features.__doc__  # noqa
```
huard, line 24: Isn't there a need for a custom docstring ?

owslib/ogcapi/edr.py
```diff
@@ -0,0 +1,98 @@
...
+        super().__init__(url, json_, timeout, headers, auth)
+
+    def data(self) -> list:
+        """
+        implements /collections filtered on EDR data resources
```
huard, line 29:
```suggestion
        Collections filtered on EDR data resources.
```

owslib/ogcapi/edr.py
```diff
@@ -0,0 +1,98 @@
...
+                if 'data' in l_['rel']:
+                    datas.append(c_['id'])
+                    break
+
+        return datas
+
+    def query_data(self, collection_id: str,
+                   query_type: str, **kwargs: dict) -> BinaryIO:
+        """
+        implements /collection/{collectionId}/coverage/
```
huard, line 48: From someone who's not familiar with the API, I suggest the docstring should start with what the function does in simple terms.

huard (approved):

owslib/ogcapi/edr.py
tomkralidis: Not at this level, given the building blocks of OGC APIs (collections are the same across APIS). docstrings are more specific deeper into the endpoints.

PR #903, opened 2024-01-30
Make GetFeature-Urls ending with & work

lordi:
This fixes a bug where a GetFeature URL ending with "&" was incorrectly constructed.

Also while debugging I was missing logs in the "owslib" namespace (as written in the docs). I think the second change in this MR fixes this.

### DIFF
--- a/owslib/feature/__init__.py
+++ b/owslib/feature/__init__.py
@@ -209,7 +209,7 @@ class WebFeatureService_(object):
                 if m.get("type").lower() == method.lower()
             )
         )
-        base_url = base_url if base_url.endswith("?") else base_url + "?"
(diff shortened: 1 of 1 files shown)


tomkralidis:
@lordi thanks for the valuable contribution.  It turns out that we log as `logging.getLogger(__name__)` in the library (though we say otherwise in the docs, README and tests).  Can you update this PR set the logger to `__name__`?  I'll update the docs/test/README in this regard for consistency.  Thanks.

lordi:
I saw that master removed owslib.util.log in the mean time, so I rebased my PR to only include the amp change.

geographika:
Closing as should be fixed in 0.32.0 release

Issue #368, opened 2017-05-17
support dimension (time) for WMTS 1.0.0


murdav:
Dear All,

Following http://portal.opengeospatial.org/files/?artifact_id=35326 there is the possibility to have the dimension, page 22 Table 9 — Parts of Dimension data structure.

e.g.
```
<Style isDefault="true">
  <ows:Title xml:lang="en">default</ows:Title>
  <ows:Identifier>default</ows:Identifier>
</Style>
<Format>image/png</Format>
<Dimension>
  <ows:Identifier>time</ows:Identifier>
  <ows:UOM>ISO8601</ows:UOM>
  <Default>2017-05-16</Default>
  <Current>false</Current>
  <Value>2015-07-30/2017-05-16/P1D</Value>
</Dimension>

```
A getcapabilities example could be:

https://gibs-b.earthdata.nasa.gov/wmts/epsg4326/best/wmts.cgi?Service=WMTS&Request=getcapabilities&Version=1.0.0

If there is no program to support it, I can have a look at it. 

@tomkralidis the issue is that the WMS-T has a different structure (page 53, 06-042_OpenGIS_Web_Map_Service_WMS_Implementation_Specification.pdf),  could you give me a bit of help, maybe some points of start?

Thanks,

D

tomkralidis:
@murdav looking at http://schemas.opengis.net/wmts/1.0/wmtsGetCapabilities_response.xsd, `wmts:Dimension` occurs 0..n times in a given `Layer` element. 

![wmts-layer-dimension](https://cloud.githubusercontent.com/assets/910430/26151901/c90fc700-3ad2-11e7-8f49-25ea85bdebc1.jpg)

 I would create a `dimensions` property which is a list of each dimension with the appropriate properties, starting at around https://github.com/geopython/OWSLib/blob/master/owslib/wmts.py#L773.  The `wmts:Value` property being a list of values.

murdav:
Thanks for this! 
Further question, I didn't find how I can run OWSLib tests before pulling the request from my fork to master. I got some minor errors (like field title has changed the value) that I can easily update. 
Is it possible to run the tests locally, if so, how?

D

tomkralidis:
@murdav you can run the tests with py.test (see how Travis does it in https://github.com/geopython/OWSLib/blob/master/.travis.yml#L17).  Note you can also run tests on single doctest files.

acerioni:
+1 for this feature request!

lubojr:
[...]

PR #904, opened 2024-02-08
add WMTS dimensions parsing

lubojr:
Implements WMTS 1.0 `domains` parsing as discussed in https://github.com/geopython/OWSLib/issues/368

Tests for WMTS are currently being skipped, but I augmented an existing test with confirmation of the time domain parsing anyway.


tomkralidis:
CI failures unrelated.

PR #921, opened 2024-04-07
Removed dependency on the pytz library.

spillner:
Most of the Pytz functionality is now available in the standard library in 3.8 and greater, and the Pytz (https://pypi.org/project/pytz/) project has gone into maintenance mode and is no longer recommended for new projects.  I noticed that OWSLib was using Pytz only to annotate timestamps as UTC, which doesn't even require the expanded tzinfo support from 3.8/3.9+; the attached patch supports Python 2 (at least Python 2.7) as well.

### DIFF
--- a/owslib/wmts.py
+++ b/owslib/wmts.py
@@ -241,8 +241,8 @@ class WebMapTileService(object):
                     parse_remote_metadata=parse_remote_metadata)
                 if cm.id:
                     if cm.id in self.contents:
-                        msg=('Content metadata for layer "%s" '
-                                       'already exists' % cm.id)
+                        msg = ('Content metadata for layer "%s" '
+                               'already exists' % cm.id)
                         warnings.warn(msg, RuntimeWarning)
                     self.contents[cm.id] = cm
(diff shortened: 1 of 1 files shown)


spillner:
Pinging the issue to prevent an auto-close.  It looks like the CI pipelines were timing out due to a test harness issue that has nothing to do with this change, and the "conflict" in requirements.txt is a fundamental part of the patch.

Anyway, LMK if there's anything else you're looking for me to do to package this up.

geographika:
Thanks @spillner for this. 
Python2 support has been dropped - and OWSLib requires Python 3.10 - https://github.com/geopython/OWSLib/blob/b9cd08b02f1c14b0e9c924ac445046f20434a590/setup.py#L92 - does this make the patch even simpler?

geographika:
@spillner - the merge conflict is due to `python-dateutil>=1.5` - the version can be removed so it looks as follows, and the CI should pass:

```
lxml
python-dateutil
pyyaml
requests
```

PR #926, opened 2024-05-12
CI: change base URL for GeoNetwork based CSW

tomkralidis:
Switch CSW GeoNetwork testing to pre-4.0 (ref: https://github.com/geonetwork/core-geonetwork/issues/6155).


tomkralidis:
CI failure is intermittent.

PR #928, opened 2024-05-26
OGC Connected Systems API

ChainReaction31:
Add support for OGC API: Connected Systems to project. Tests are run against OpenSensorhub's current implementation of the api.

Work is an extension of the current OGC API module and extends from Collections.

Some changes deviate from what was set forth in that effort to reduce the number of requests needed to be made to retrieve things like IDs that are return as part of the header files upon successful resource creation. A similar change was made to updates and deletions to give a little more feedback to a user when these operations are successful.

This work was done in service of adding another reference implementation of Connected Systems API and to hopefully get it in the hands of an already established user base.

- Ian Patterson (ChainReaction31)


tomkralidis:
CI failures are unrelated.

tomkralidis:
Merging, great work @ChainReaction31 !

PR #930, opened 2024-06-01
remove root loggers (#923)

### LINKED ISSUE #923: Logging module references inconsistent and unpracticable for configuration with other packages
Some places use these references to obtain a logger: 
https://github.com/geopython/OWSLib/blob/79909b497cdcc691899246368b622bc7fec15c4f/owslib/util.py#L10
https://github.com/geopython/OWSLib/blob/79909b497cdcc691899246368b622bc7fec15c4f/owslib/wps.py#L115-L116
https://github.com/geopython/OWSLib/blob/79909b497cdcc691899246368b622bc7fec15c4f/owslib/wps.py#L308

This causes the `log.<level>` function called to use the `root` logger.

On the other hand, the following use a specific logger (resulting in `owslib.wps`):
https://github.com/geopython/OWSLib/blob/79909b497cdcc691899246368b622bc7fec15c4f/owslib/wps.py#L144

Because of this, there is no clean way to configure the logger all over `owslib`. 
Even worst is that, when used inside another application, the `root` logger must often be silenced or raised to `ERROR` or `WARNING` level to avoid too much noise, since it can combine many other locations. This makes it unpractical to debug specific operations in `owslib`. 

The logging references should be unified to all employ (at least) the `owslib` namespace instead of `root`.
A quick fix would be to replace `import logging as log` in `owslib.util` by the following: 
```python
import logging
log = logging.getLogger(__name__)
```
This way, any place "incorrectly" using the `logging` alias as `log` would at least transparently use a namespace'd logger instead. 

Maybe there are other locations to patch, but I have not explored the full repository for other logging references.

tomkralidis:
Fixes #923

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,19 +127,19 @@ Map Tile Service (WMTS). Some of those are beta quality.
 
 Logging
 -------
-OWSLib logs messages to the 'owslib' named Python logger. You may
+OWSLib logs messages to module-based named Python loggers. You may
 configure your application to use the log messages like so:
 
 ```python
     >>> import logging
-    >>> owslib_log = logging.getLogger('owslib')
+    >>> LOGGER = logging.getLogger(__name__)
     >>> # Add formatting and handlers as needed, for example to log to the console
     >>> ch = logging.StreamHandler()
     >>> ch.setLevel(logging.DEBUG)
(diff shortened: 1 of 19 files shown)


tomkralidis:
Test failures are unrelated.

cehbrecht (approved):
Looks good to me. Cleans up the code :)

PR #931, opened 2024-06-01
OGC API: capture response headers

tomkralidis:


tomkralidis:
Test failures are unrelated/intermittent.

kalxas (approved):

PR #941, opened 2024-09-29
drop Python 3.8/3.9, unpin old dependencies

tomkralidis:


kalxas (approved):

gamesbook:
Python 3.9 is still officially supported for another year; any key reason to drop it now?

Issue #910, opened 2024-03-09
[WFS 2.0.0] getfeature argument to specify SRS/CRS?


digital-idiot:
As of version `'0.29.2'` there is no way to specify CRS/SRS when calling `WebFeatureService_2_0_0.getfeature` .
See:

https://github.com/geopython/OWSLib/blob/f8e3ff65ad70d72a640e2b7883459b8aa8fcb4b6/owslib/feature/wfs200.py#L220-L287)

However both in `WebFeatureService_1_1_0` and `WebFeatureService_1_0_0` have the following parameter to specify SRS:

```
srsname: string 
             EPSG code to request the data in 
```

https://github.com/geopython/OWSLib/blob/f8e3ff65ad70d72a640e2b7883459b8aa8fcb4b6/owslib/feature/wfs110.py#L216-L269

https://github.com/geopython/OWSLib/blob/f8e3ff65ad70d72a640e2b7883459b8aa8fcb4b6/owslib/feature/wfs100.py#L208-L255

jnicho02:
ditto #905

lukas-paulun:
I would also be interested in an answer to that question

geographika:
Fixed with #945

Issue #905, opened 2024-02-15
WFS 2.0.0 getFeature srsname


jnicho02:
In WFS 2.0.0 there is no srsname parameter in the getfeature call.

Comparing [wfs110.py](https://github.com/geopython/OWSLib/blob/master/owslib/feature/wfs110.py) to [wfs200.py](https://github.com/geopython/OWSLib/blob/master/owslib/feature/wfs200.py) it doesn't seem hard to implement, but was it intentionally left out?

jnicho02:
I sincerely hope that the PR that fixes this small, yet significant, problem will make it through. This issue forced me to delist OWSLib from the set of libraries our Devs can use.

geographika:
Fixed with #945

Issue #682, opened 2020-05-01
parameter srsname is missing  in .getfeature  


JoukeVerstappen:

iherwig:
To be more specific, `WebFeatureService_2_0_0` is missing the `srsname` argument in the `getfeature` method. Is it missing by purpose?

PR #945, opened 2024-09-30
Add srsname to wfs200.py getfeature GET request

lpartington:
WebFeatureService_2_0_0.getfeature() is missing an srsName argument. This issue has been reported previously in #910, #905, and #682. This prevents features being retrieved with a non-default crs when using WFS 2.0.0. srsName _is_ supported by WebFeatureService_1_1_0.getfeature() and WebFeatureService_1_0_0.getfeature(). This seems to be an oversight as srsname is a valid KVP ('key value pair') parameter - see OGC WFS 2.0.2 specification https://portal.ogc.org/files/?artifact_id=39967 Sections 7.9.2.3 and 7.9.2.4.4.

Changes in this PR:
* Add srsname arg to WebFeatureService_2_0_0.getfeature() for GET requests
* Add srsname arg to WebFeatureService_.getGETGetFeatureRequest() (which is called by getfeature)

Note that srsname hasn't been added for POST requests. This is actually consistent with the implementation of getfeature() in wfs110.py, where srsname can be specified for GET requests but not POST requests. It seems this is likely to be unintentional, but fixing up POST requests across all WFS versions is better done in a separate PR.

A suggestion for future improvement would be to support kwargs in getfeature(), so that arbitrary query parameters can be supported without having to explicitly declare them in the function argument list. These parameters would merely be added verbatim to the request's query params. However, this is beyond the scope of the current issue.

### DIFF
--- a/owslib/feature/__init__.py
+++ b/owslib/feature/__init__.py
@@ -160,6 +160,7 @@ class WebFeatureService_(object):
         featureversion=None,
         propertyname=None,
         maxfeatures=None,
+        srsname=None,
         storedQueryID=None,
         storedQueryParams=None,
         outputFormat=None,
@@ -183,6 +184,8 @@ class WebFeatureService_(object):
             List of feature property names. '*' matches all.
         maxfeatures : int
             Maximum number of features to be returned.
+        srsname: string
+            EPSG code to request the data in
         method : string
             Qualified name of the HTTP DCP method to use.
         outputFormat: string (optional)
@@ -238,6 +241,13 @@ class WebFeatureService_(object):
                 request["count"] = str(maxfeatures)
             else:
                 request["maxfeatures"] = str(maxfeatures)
+        if srsname:
+            request["srsname"] = str(srsname)
+
+            # Check if desired SRS is supported by the service for each
+            # typename. Warning will be thrown if that SRS is not allowed.
+            for name in typename:
+                _ = self.getSRS(srsname, name)
         if startindex:
             request["startindex"] = str(startindex)
         if storedQueryID:
--- a/owslib/feature/wfs200.py
+++ b/owslib/feature/wfs200.py
@@ -226,6 +226,7 @@ class WebFeatureService_2_0_0(WebFeatureService_):
         featureversion=None,
         propertyname=None,
         maxfeatures=None,
+        srsname=None,
         storedQueryID=None,
         storedQueryParams=None,
         method="Get",
@@ -254,6 +255,8 @@ class WebFeatureService_2_0_0(WebFeatureService_):
             For Post request, leave blank (None) to get all properties.
         maxfeatures : int
             Maximum number of features to be returned.
+        srsname: string
+            EPSG code to request the data in
         storedQueryID : string
             A name identifying a prepared set available in WFS-service
         storedQueryParams : dict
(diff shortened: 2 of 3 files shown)


geographika:
@lpartington - would you be able to add a test for this? Some basic tests of `getGETGetFeatureRequest` with `srsName` should be fine.

lpartington:
No problem. I will look at it tomorrow.

lpartington:
I added test case test_srsname_wfs_200 in test_wfs_generic.py, using the same approach as the wfs_100 and wfs_110 cases. However, I noted these issues:

1. most of the cases in test_wfs_generic.py depend on wfs service https://www.sciencebase.gov/catalog/item/53398e51e4b0db25ad10d288, but it appears that sciencebase services don't currently work. In all examples I tried, an ows:ExceptionReport was returned with 
`<ows:ExceptionReport xmlns:ows="http://www.opengis.net/ows" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.0.0" xsi:schemaLocation="http://www.opengis.net/ows https://www.sciencebase.gov:443/catalogMaps/schemas/ows/1.0.0/owsExceptionReport.xsd">
<script/>
    <ows:Exception exceptionCode="NoApplicableCode">
        <ows:ExceptionText>java.lang.NoClassDefFoundError: Lorg/eclipse/core/resources/IWorkspaceRoot; Lorg/eclipse/core/resources/IWorkspaceRoot; org.eclipse.core.resources.IWorkspaceRoot</ows:ExceptionText>
    </ows:Exception>
</ows:ExceptionReport>
`
The WMS services are also not working (but returning a slightly different ExceptionReport). 
I have changed all of the 'online' tests in test_wfs_generic.py to use the service that was used in the test_xmlfilter_wfs_xxx cases - https://services.ga.gov.au/gis/stratunits/ows. All but one of the 'online' cases are now passing.
2. All of the 'online' cases in test_wfs_generic.py were marked with xfail. I've removed xfail from all cases that are now passing.
3. The test case that is failing is test_srsname_wfs_110. The request for EPSG:99999999 is not raising a ServiceException. It seems that there is an issue with how openUrl and the getfeature implementations are handling errors for WFS 1.1.0. 

For the wfs_110 case, the service returns 200 OK with an ows:ExceptionReport body. It looks like the intention of openURL is to catch such errors and raise a ServiceException in the possible_errors code, but this doesn't occur because the Content-Type of the response is "application/xml;charset=utf-8" rather than "application/xml". I wonder if substrings starting with ';' should be removed from the content type prior to the comparison?

Note that the wfs_100 and wfs_200 cases DO raise ServiceException for this EPSG. This is because the service returns 400 Bad Request (it doesn't seem right that the service returns different status codes for the different WFS versions, but it does!) and openUrl raising ServiceException as a result.
4. It is worth mentioning that the GA stratunits feature type doesn't have any geometry, so updated the test_schema_wfs_xxx cases to verify that geometry is None. This doesn't seem ideal; perhaps a different service/feature type should be used that has geometry.

geographika:
Thanks @lpartington for adding those tests, and taking the time to get the WFS tests working again. 
The CI is failing due to coveralls.io issues. If you're able to rebase this against the latest master it may fix it. 

+1 from me on merging this in the next week, unless there are any other comments from others.

geographika:
CI all passing. Thanks @lpartington !

PR #949, opened 2024-10-08
Remove dependency on the Pytz library

spillner:
Pytz no longer supports the latest Python releases, and not much of it was actually used by OWSLib anyway.

Recreated the previous PR #921 to permit branching my source repository, to better manage the rebase and elide false-positive merge conflicts reported by GitHub.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,20 +7,18 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.10", "3.11"]
     env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         COVERALLS_SERVICE_NAME: github
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-python@v2
+    - uses: actions/checkout@master
+    - uses: actions/setup-python@v5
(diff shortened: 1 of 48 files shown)


geographika:
Thanks @spillner for updating the pull request.
Would it be possible to add a very simple test to test_util.py to parse a date using the new TimeZone_UTC class and check the output?

I'm +1 on merging once that is in place.

geographika:
Tested the branch locally - all working as expected:

```python
>>> import pytz
>>> from datetime import datetime, timedelta
>>> dt = datetime.utcnow()
>>> dt.replace(tzinfo=pytz.utc)
datetime.datetime(2024, 10, 8, 10, 7, 59, 167706, tzinfo=<UTC>)
>>> from owslib.util import TimeZone_UTC
>>> datetime.utcnow().replace(tzinfo=TimeZone_UTC())
datetime.datetime(2024, 10, 8, 10, 9, 31, 262084, tzinfo=<owslib.util.TimeZone_UTC object at 0x0000015D9D187CD0>)
```

spillner:
Added two unit tests, one for datetime normalization as a standalone operation, and one for proper operation of extract_time() as a whole.

I also took advantage of the shift to 3.10 and newer to drop the TimeZone_UTC class in favor of the standard datetime.timezone.utc.  If anyone needs to restore support for 3.8 and earlier, you could just roll back the very last commit in this series.

geographika:
@spillner - thanks very much for this. Removing a dependency and being able to use the standard library is excellent.

Issue #951, opened 2024-10-25
`tests/test_ogcapi_connectedsystems_osh.py` requires Internet but not marked as `online`


mgorny:
When running the test suite in a container isolated from the Internet, with `-m "not online"` option, it started failing in 0.32.0 release:

```pytb
========================================================= test session starts =========================================================
platform linux -- Python 3.12.7, pytest-8.3.3, pluggy-1.5.0 -- /tmp/portage/dev-python/owslib-0.32.0/work/OWSLib-0.32.0-python3_12/install/usr/bin/python3.12
cachedir: .pytest_cache
rootdir: /tmp/portage/dev-python/owslib-0.32.0/work/OWSLib-0.32.0
configfile: tox.ini
collecting ... collected 239 items / 1 error / 71 deselected / 168 selected

=============================================================== ERRORS ================================================================
_____________________________________ ERROR collecting tests/test_ogcapi_connectedsystems_osh.py ______________________________________
/usr/lib/python3.12/site-packages/urllib3/connection.py:199: in _new_conn
    sock = connection.create_connection(
        self       = <urllib3.connection.HTTPConnection object at 0x7f0d8dbed9a0>
/usr/lib/python3.12/site-packages/urllib3/util/connection.py:85: in create_connection
    raise err
        address    = ('34.67.197.57', 8585)
        af         = <AddressFamily.AF_INET: 2>
        canonname  = ''
        err        = None
        family     = <AddressFamily.AF_UNSPEC: 0>
        host       = '34.67.197.57'
        port       = 8585
        proto      = 6
        res        = (<AddressFamily.AF_INET: 2>,
 <SocketKind.SOCK_STREAM: 1>,
 6,
 '',
 ('34.67.197.57', 8585))
        sa         = ('34.67.197.57', 8585)
        sock       = <socket.socket [closed] fd=-1, family=2, type=1, proto=6>
        socket_options = [(6, 1, 1)]
        socktype   = <SocketKind.SOCK_STREAM: 1>
        source_address = None
        timeout    = None
/usr/lib/python3.12/site-packages/urllib3/util/connection.py:73: in create_connection
    sock.connect(sa)
E   OSError: [Errno 101] Network is unreachable
        address    = ('34.67.197.57', 8585)
        af         = <AddressFamily.AF_INET: 2>
        canonname  = ''
        err        = None
        family     = <AddressFamily.AF_UNSPEC: 0>
        host       = '34.67.197.57'
        port       = 8585
[...]

PR #952, opened 2024-10-27
Add offline testing to CI

geographika:
This pull request makes use of the https://github.com/miketheman/pytest-socket/ plugin to check the test suite runs without network access. Relates to #951. 

There is no simple way to mark doctests as requiring online access, or to disable the `--doctest-modules` flag from the command line, so I have manually deselected the 3 tests that require network access in the main.yml file. 

To ensure the test suite passes additional `@pytest.mark.online` have been added to tests requiring network access. 

`tox.ini` was modified to put options on separate lines to improve readibility.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -27,6 +27,15 @@ jobs:
         pip3 install -r docs/requirements.txt
     - name: run tests ⚙️
       run: python3 -m pytest
+    - name: run tests in offline mode
+      if: matrix.python-version == '3.10'
+      run: |
+          python3 -m pytest \
+            -m "not online" \
+            --disable-socket \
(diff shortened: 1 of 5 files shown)


geographika:
From comments in the merged pull request from @mgorny (to avoid these details being lost):

> Move the OSHFixtures class into a pytest fixture to defer initializing it (and therefore accessing the Internet) until the relevant test functions are called. This makes it possible for pytest.mark.online to correctly skip them. It also avoids initializing the fixtures class multiple times -- now it should be initialized once per session.

> While at it, move the global functions into the OSHFixtures class. It would need to access the class members anyway, and this seems cleaner than passing its instance explicitly.

Issue #878, opened 2023-06-23
Bug in bounding box creation for WMS 1.1.1


om-henners:
Currently this line causes an issue when building the URL parameter for the bounding box in WMS 1.1.1

https://github.com/geopython/OWSLib/blob/bc88ee1a6509746cff1cccfbad6e6501bae4c256/owslib/map/wms111.py#L175

and likewise in WMS 1.3.0, although only when the projection for the bounding box is in a non-geographic projection (because above if in a geographic projection the bbox is remapped to reverse the axis).

https://github.com/geopython/OWSLib/blob/bc88ee1a6509746cff1cccfbad6e6501bae4c256/owslib/map/wms130.py#L186

The result is that it adds quotes into the parameter string which causes downstream issues (or at least does in geoserver). For example a bounding box of 

```
bbox = (0, 0, 1, 1, 'EPSG:28356')
```

is reformatted as `0,0,1,1,'EPSG:28356'` when the server is expecting `0,0,1,1,EPSG:28356`.

The solution is to use `str` instead of `repr` for the join.

geographika:
Fixed with #955

PR #955, opened 2024-11-06
FIX: numpy floats parsed incorrectly

juhi24:
Referring to https://github.com/SciTools/cartopy/issues/2472#issuecomment-2457518215

Currently `owslib` cannot handle bbox defined as `numpy.float64`. This PR fixes the referred issue by using `str` instead of `repr`.

### DIFF
--- a/owslib/coverage/wcs110.py
+++ b/owslib/coverage/wcs110.py
@@ -176,7 +176,7 @@ class WebCoverageService_1_1_0(WCSBase):
         request['identifier'] = identifier
         # request['identifier'] = ','.join(identifier)
         if bbox:
-            request['boundingbox'] = ','.join([repr(x) for x in bbox])
+            request['boundingbox'] = ','.join([str(x) for x in bbox])
         if time:
             request['timesequence'] = ','.join(time)
         request['format'] = format
--- a/owslib/feature/wfs100.py
+++ b/owslib/feature/wfs100.py
@@ -268,7 +268,7 @@ class WebFeatureService_1_0_0(object):
         if featureid:
             request["featureid"] = ",".join(featureid)
         elif bbox and typename:
-            request["bbox"] = ",".join([repr(x) for x in bbox])
+            request["bbox"] = ",".join([str(x) for x in bbox])
         elif filter and typename:
             request["filter"] = str(filter)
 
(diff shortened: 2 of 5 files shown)


geographika:
Thanks @juhi24. The CI looks like it is failing on unrelated issues. 

It looks like this line was originally `request['bbox'] = ','.join([str(x) for x in bbox])` but was updated to use `repr` in 
https://github.com/geopython/OWSLib/commit/c9e2ddb426683f8b0036e370826906a8030241c5 - the reason being "changing str to repr to maintain precision in bbox coords".

It would be good to add a test with a high precision bbox values to check if `str` strips values off the end. 

Also linking to a related issue #878.

juhi24:
Hi, @geographika. I cannot reproduce the difference in precision in the earliest and latest supported Python versions 3.10 or 3.13, nor in legacy 3.6. All give:

```python
>>> f = 123.4567890123456789
>>> str(f)
'123.45678901234568'
>>> repr(f)
'123.45678901234568'
```

I think this was a Python 2 issue:
```python
>>> f = 123.4567890123456789
>>> str(f)
'123.456789012'
>>> repr(f)
'123.45678901234568'
```

The https://github.com/geopython/OWSLib/commit/c9e2ddb426683f8b0036e370826906a8030241c5 was committed in 2011.

geographika:
@juhi24 - thanks for checking this, I'm +1 on merging. 

It would be good to add in a test just confirming the precision for a bbox is retained. I think adding a simple unit test for `__build_getmap_request` (for WMS 1.3.0) to `test_wms_getmap.py` would work, without needing to make an actual request to a server. If you're able to add this to the pull request that would be great.

juhi24:
@geographika, test added and passing.

geographika:
Just for future reference a float is truncated when converting with `str`, but in Python3 it is rounded to the same number of decimal places as `repr` so functionality will be unchanged with this pull request. To have higher accuracy use decimals - this approach would not have been possible using `repr`. 

```python
>>> f = 123.4567890123456789
>>> str(f)
'123.45678901234568'
>>> repr(f)
'123.45678901234568'

# to have higher precision use decimals - str converts correctly
>>> str(d)
'123.4567890123456805895330035127699375152587890625'
>>> repr(d)
"Decimal('123.4567890123456805895330035127699375152587890625')"
# this would fail to create a bbox
```

Thanks for the test and fix @juhi24.

Issue #956, opened 2024-11-07
WFS GetFeature filter tests currently failing


geographika:
2 tests in the CI are currently failing: `test_xmlfilter_wfs_110` and `test_xmlfilter_wfs_200`. These both use the server https://services.ga.gov.au/gis/stratunits/ows

The server is still online, and requests such as https://services.ga.gov.au/gis/stratunits/wfs?service=WFS&version=1.1.0&request=GetCapabilities and https://services.ga.gov.au/gis/stratunits/wfs?service=WFS&version=1.1.0&request=GetFeature&typename=stratunit%3AStratigraphicUnit&maxfeatures=2 both return valid data. 

It appears that adding a `filter` returns `403 Forbidden` responses. 

The feature attributes have changed as the tests are expecting a `GEOLOGICHISTORY` and `StratigraphicUnit` attributes. The new fields are below. Changing the filter to use these new fields however still returns `403 Forbidden` responses. 

```xml
  <gml:featureMember>
    <stratunit:StratigraphicUnit gml:id="StratigraphicUnit.5234">
      <stratunit:STRATNO>5234</stratunit:STRATNO>
      <stratunit:GMLID>ga.gsmlp.geologicunitview.5234</stratunit:GMLID>
      <stratunit:IDENTIFIER>http://pid.geoscience.gov.au/geologicFeature/au/SU5234</stratunit:IDENTIFIER>
      <stratunit:NAME>Darriwil granite</stratunit:NAME>
      <stratunit:DESCRIPTION>Granite</stratunit:DESCRIPTION>
      <stratunit:GEOLOGICUNITTYPE>lithostratigraphic unit</stratunit:GEOLOGICUNITTYPE>
      <stratunit:RANK>Formation, beds</stratunit:RANK>
      <stratunit:LITHOLOGY>granite</stratunit:LITHOLOGY>
      <stratunit:OBSERVATIONMETHOD>synthesis from multiple sources</stratunit:OBSERVATIONMETHOD>
      <stratunit:GEOLOGICUNITTYPE_URI>http://resource.geosciml.org/classifier/cgi/geologicunittype/lithostratigraphic_unit</stratunit:GEOLOGICUNITTYPE_URI>
      <stratunit:REPRESENTATIVELITHOLOGY_URI>http://resource.geosciml.org/classifier/cgi/lithology/granite</stratunit:REPRESENTATIVELITHOLOGY_URI>
[...]

Issue #936, opened 2024-07-09
properly handle 403 return codes from remote services


landryb:
using 0.27.2 on debian 12, in some cases i'm creating a `WebMapService` object with an url that requires authentication, but i don't know that it requires auth yet. The object creation fails with a traceback related to lxml/etree:
```
Traceback (most recent call last):
  File "/home/landry/scratch/testmsorm/test.py", line 134, in <module>
    s = WebMapService(u, version='1.3.0')
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/owslib/wms.py", line 54, in WebMapService
    return wms130.WebMapService_1_3_0(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/owslib/map/wms130.py", line 75, in __init__
    self._capabilities = reader.read(self.url, timeout=self.timeout)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/owslib/map/common.py", line 69, in read
    return etree.fromstring(raw_text)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "src/lxml/etree.pyx", line 3257, in lxml.etree.fromstring
  File "src/lxml/parser.pxi", line 1916, in lxml.etree._parseMemoryDocument
  File "src/lxml/parser.pxi", line 1803, in lxml.etree._parseDoc
  File "src/lxml/parser.pxi", line 1144, in lxml.etree._BaseParser._parseDoc
  File "src/lxml/parser.pxi", line 618, in lxml.etree._ParserContext._handleParseResultDoc
  File "src/lxml/parser.pxi", line 728, in lxml.etree._handleParseResult
  File "src/lxml/parser.pxi", line 657, in lxml.etree._raiseParseError
  File "<string>", line 13
lxml.etree.XMLSyntaxError: Opening and ending tag mismatch: meta line 11 and head, line 13, column 8
```
i can of course catch that exception and handle it as "something failed" but in that case the server just returned a 403 code which is .. not handled by https://github.com/geopython/OWSLib/blame/master/owslib/util.py#L210

if i add 403 to that line, an `owslib.util.ServiceException` is raised. Should it be added to the list of codes ?

[...]

PR #958, opened 2024-11-07
Catch and raise HTTP 403 Forbidden errors

geographika:
It would have been easier to see the reason for recent failing CI tests (see #956) if an error had been raised when HTTP code 403 was returned. 

As HTTP error code is a client error code, I added to the list with 400 and 401 (rather than the server error codes 5xx list). 

See also #936. 

Note, this could break existing workflows by raising a ServiceException - so comments/thoughts welcome.


landryb:
matches what i had tried in #936, much welcomed !

PR #963, opened 2024-12-03
capture dct:spatial as bbox, if it contains a box

pvgenuchten:
in some cases, bbox is strored in dct:spatial, nice to have it as bbox

added some csw parse tests, seemed not existing


geographika:
@pvgenuchten - great to get some tests for the CSW parsing. It looks like there is a lot of shared code in the `__init__` for `CswRecord` and `Csw30Record` that could be refactored into one function - maybe for another pull request. 

A couple of questions:

1. If a document has both `ows:BoundingBox` and `dct:spatial` then `self.bbox` will overwrite the value from `ows:BoundingBox` with  `dct:spatial`. Is this scenario possible, and if yes then should this be the correct precedence?
2. If `dct:spatial` is not a bbox in the form `-180,-90,180,90` then `self.spatial` is set to `val`. What could `val` be - a string? It would be good to have an example document and test for this.

geographika:
@pvgenuchten - we're looking to make a new release of OWSLib. If you have any thoughts on the above questions, please add to this ticket. Are you then happy to merge?

pvgenuchten:
> @pvgenuchten - great to get some tests for the CSW parsing. It looks like there is a lot of shared code in the `__init__` for `CswRecord` and `Csw30Record` that could be refactored into one function - maybe for another pull request.

I'm not sure i'm able to manage the refactore, but for sure is a good idea. I'd say postpone to next PR.

> 1. If a document has both `ows:BoundingBox` and `dct:spatial` then `self.bbox` will overwrite the value from `ows:BoundingBox` with  `dct:spatial`. Is this scenario possible, and if yes then should this be the correct precedence?

valid question, I will have a look, maybe we can store both?

> 2. If `dct:spatial` is not a bbox in the form `-180,-90,180,90` then `self.spatial` is set to `val`. What could `val` be - a string? It would be good to have an example document and test for this.

dct:spatial indeed is often populated with a spatial label, such as `London`, `Bretagne`. Indeed good to add a test for it.

tests/test_csw_parsing.py
```diff
@@ -0,0 +1,69 @@
...
+    """Test the parsing of a metadatarecord 
+
+    GetRecordById response available in
+    tests/resources/9250AA67-F3AC-6C12-0CB9-0662231AA181_dc2.xml
+
+    """
+    md_resource = get_md_resource('tests/resources/9250AA67-F3AC-6C12-0CB9-0662231AA181_dc2.xml')
+    md = CswRecord(md_resource)
+
+    assert type(md) is CswRecord
```
tomkralidis:
```python
assert isinstance(CswRecord, md)
```

tests/test_csw_parsing.py
```diff
@@ -0,0 +1,69 @@
...
+
+    Returns
+    -------
+    etree.ElementTree
+        XML tree of the resource on disk.
+
+    """
+    namespaces = Namespaces().get_namespaces(keys=('dc', 'dct', 'ows', 'rdf', 'gml', 'csw'))
+
+    with io.open(file_path, mode='r', encoding='utf-8') as f:
```
tomkralidis: Can we use `open()` instead of `io.open`?

owslib/catalogue/csw2.py
```diff
@@ -849,15 +849,25 @@ def __init__(self, record):
...
-        val = record.find(util.nspath_eval('dct:spatial', namespaces))
-        self.spatial = util.testXMLValue(val)
-
         val = record.find(util.nspath_eval('ows:BoundingBox', namespaces))
         if val is not None:
             self.bbox = ows.BoundingBox(val, namespaces['ows'])
         else:
             self.bbox = None
 
+        val = record.find(util.nspath_eval('dct:spatial', namespaces))
```
tomkralidis: Suggest the below (also for `owslib/catalogue/csw3.py`):
```python
        val = record.find(util.nspath_eval('dct:spatial', namespaces))
        self.spatial = None
        if val is not None:
            val = util.testXMLValue(val)
            tokens = val.split(',')
            if len(tokens) == 4:
                self.bbox = ows.BoundingBox(None, namespaces['ows'])
                self.bbox.minx = tokens[0]
                self.bbox.miny = tokens[1]
                self.bbox.maxx = tokens[2]
                self.bbox.maxy = tokens[3]
            else:
                self.spatial = val
```

tomkralidis (changes requested):

owslib/catalogue/csw2.py
pvgenuchten: it seems already there

pvgenuchten:
wfs test failures not related

owslib/catalogue/csw2.py
tomkralidis, line 858:
```python
        val = record.find(util.nspath_eval('dct:spatial', namespaces)) 1Code has comments. Press enter to view.
        self.spatial = None
        if val is not None:
            val = util.testXMLValue(val)
            bbox_tokens = val.split(',')
            if len(bbox_token) == 4:
                self.bbox = ows.BoundingBox(None, namespaces['ows'])
                self.bbox.minx = bbox_tokens[0]
                self.bbox.miny = bbox_tokens[1]
                self.bbox.maxx = bbox_tokens[2]
                self.bbox.maxy = bbox_tokens[3]
            else:
                self.spatial = val
```

owslib/catalogue/csw3.py
```diff
@@ -757,6 +757,19 @@ def __init__(self, record):
         else:
             self.bbox = None
 
+        val = record.find(util.nspath_eval('dct:spatial', namespaces))
+        self.spatial = None
+        if val is not None:
```
tomkralidis, line 762:
```python
        if val is not None:
            val = util.testXMLValue(val)
            bbox_tokens = val.split(',')
            if len(bbox_token) == 4:
                self.bbox = ows.BoundingBox(None, namespaces['ows'])
                self.bbox.minx = bbox_tokens[0]
                self.bbox.miny = bbox_tokens[1]
                self.bbox.maxx = bbox_tokens[2]
                self.bbox.maxy = bbox_tokens[3]
            else:
                self.spatial = val
```

PR #967, opened 2025-01-16
fix OGC API coverages `subset` and `datetime` not forwarded

fmigneault:
Fixes needed for OGC TB-20 TIEs. 

Some parameters are not forwarded to the coverage endpoint, causing requests to be refused by servers that consider the subset to be too large to return reasonably.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,6 @@ tags
 tester.py
 tags
 .idea/
-.run/
 tile.jpg
 /env/
 /.project
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ tags
 tester.py
 tags
 .idea/
+.run/
 tile.jpg
 /env/
 /.project
--- a/owslib/ogcapi/coverages.py
+++ b/owslib/ogcapi/coverages.py
@@ -58,9 +58,9 @@ class Coverages(Collections):
         @param scale_factor: factor by which to scale the resulting coverage
         @type scale_axes: list of tuples
         @param scale_axes: [(axis name, number)]
-        @type datetime: tuple | str
+        @type datetime: tuple | list | str
         @param datetime:
-            tuple of start/end datetimes or as 'start/end' string
+            tuple or list of start/end datetimes, or as 'start/end' string
             start and end datetimes can be ".." for unbounded value
 
         @returns: coverage data
@@ -89,8 +89,8 @@ class Coverages(Collections):
         if 'scale_factor' in kwargs:
             kwargs_['scale-factor'] = int(kwargs['scale_factor'])
 
-        if "datetime" in kwargs:
-            if isinstance(kwargs['datetime'], tuple):
+        if 'datetime' in kwargs:
+            if isinstance(kwargs['datetime'], (tuple, list)):
                 kwargs_['datetime'] = '/'.join(kwargs['datetime'][:2])
             else:
                 kwargs_['datetime'] = str(kwargs['datetime'])
--- a/owslib/ogcapi/coverages.py
+++ b/owslib/ogcapi/coverages.py
@@ -58,6 +58,10 @@ class Coverages(Collections):
         @param scale_factor: factor by which to scale the resulting coverage
         @type scale_axes: list of tuples
         @param scale_axes: [(axis name, number)]
+        @type datetime: tuple | str
+        @param datetime:
+            tuple of start/end datetimes or as 'start/end' string
+            start and end datetimes can be ".." for unbounded value
 
         @returns: coverage data
         """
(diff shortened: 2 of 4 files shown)


owslib/ogcapi/coverages.py
```diff
@@ -58,6 +58,10 @@ def coverage(self, collection_id: str, **kwargs: dict) -> BinaryIO:
         @param scale_factor: factor by which to scale the resulting coverage
         @type scale_axes: list of tuples
         @param scale_axes: [(axis name, number)]
+        @type datetime: tuple | str
```
tomkralidis, line 61: Can we support a `list` as well?  Would be easier for a user too.

owslib/ogcapi/coverages.py
```diff
@@ -80,11 +84,17 @@ def coverage(self, collection_id: str, **kwargs: dict) -> BinaryIO:
...
             for s in kwargs['subset']:
                 subsets_list.append(f'{s[0]}({s[1]}:{s[2]})')
-            kwargs['subset'] = ','.join(subsets_list)
+            kwargs_['subset'] = ','.join(subsets_list)
 
         if 'scale_factor' in kwargs:
             kwargs_['scale-factor'] = int(kwargs['scale_factor'])
 
+        if "datetime" in kwargs:
+            if isinstance(kwargs['datetime'], tuple):
```
tomkralidis, line 93:
```python
if isinstance(kwargs['datetime'], (list, tuple)):
```

tomkralidis (changes requested):

geographika:
@fmigneault @tomkralidis - the failing test is due to making an external request in the offline testing. Adding the decorator `@pytest.mark.online` to `test_ogcapi_build_url` should fix this:

https://github.com/geopython/OWSLib/blob/8c7effc7a422a32753579cb56b731fca7fd08a8a/tests/test_ogcapi_records_pycsw.py#L76

It looks like creating any OGC API class will make an HTTP directly on `__init__` - I think this differs from WMS/WFS where the class can be created and then HTTP requests made with a different call.

https://github.com/geopython/OWSLib/blob/8c7effc7a422a32753579cb56b731fca7fd08a8a/owslib/ogcapi/__init__.py#L68

tomkralidis:
> @fmigneault @tomkralidis - the failing test is due to making an external request in the offline testing. Adding the decorator `@pytest.mark.online` to `test_ogcapi_build_url` should fix this:
> 
> https://github.com/geopython/OWSLib/blob/8c7effc7a422a32753579cb56b731fca7fd08a8a/tests/test_ogcapi_records_pycsw.py#L76
> 

+1

> It looks like creating any OGC API class will make an HTTP directly on `__init__` - I think this differs from WMS/WFS where the class can be created and then HTTP requests made with a different call.
> 
> https://github.com/geopython/OWSLib/blob/8c7effc7a422a32753579cb56b731fca7fd08a8a/owslib/ogcapi/__init__.py#L68

Not sure. WMS support always does a GetCapabilities on init.

owslib/ogcapi/coverages.py
fmigneault: Yes. No problem. 

I've also noticed that `subset: [(name, lower bound, upper bound)]` are often provided as `dict[str, tuple[lower bound, upper bound]]` (more similar to typical JSON representation). 

I could add this as well while at it.

tomkralidis: Sure.

fmigneault:
I've added more tests (all passed) for the various kw updates related to OGC API coverages.
Failing tests are unrelated to this PR (records and csw tests).

.gitignore
```diff
@@ -21,6 +21,7 @@ tags
 tester.py
 tags
 .idea/
+.run/
```
tomkralidis, line 24: Can this be removed?

tomkralidis (changes requested):

.gitignore
fmigneault: I can remove it, but this is the new location that JetBrains IDE favor to store configurations instead of `.idea/`, so they might be introduced by someone unintentionally.

fmigneault:
@tomkralidis 
Thanks for merging. 
Are there plans for a new release in a near future?

tomkralidis:
> @tomkralidis Thanks for merging. Are there plans for a new release in a near future?

Probably a good time for 0.31.1 then (@geographika thoughts?).

geographika:
@tomkralidis @fmigneault - yes, I can make a new release this week. There are quite a few changes, but all appear so far to be bug fixes, so it would be 0.32.1 - although if #963 is merged I think it would bump it to 0.33.0 as it is a change in functionality.

tomkralidis:
> @tomkralidis @fmigneault - yes, I can make a new release this week. There are quite a few changes, but all appear so far to be bug fixes, so it would be 0.32.1 - although if #963 is merged I think it would bump it to 0.33.0 as it is a change in functionality.

Sorry, typo, yes 0.32.1.

fmigneault:
Hi @geographika 
Doing a follow-up about this. The other PR doesn't seem to be receiving the requested feedback. Is it being worked on or should this be released separately? 
Thanks

geographika:
@fmigneault - no #963 not yet merged, but this pull request has been released in https://github.com/geopython/OWSLib/releases/tag/0.32.1 and is available at https://pypi.org/project/OWSLib/
You should have got a ping from the release as you're tagged (and also posted to the mailing lists https://lists.osgeo.org/mailman/listinfo/owslib-users and https://lists.osgeo.org/mailman/listinfo/owslib-devel).

fmigneault:
@geographika  Thanks

Issue #965, opened 2025-01-14
OperationsMetadata allows invalid nameless element


CJGutz:
There is an issue when instantiating the `WebMapTileService` object in some edge cases. Looping through operation metadata, it does not exclude metadata that lies under "OperationsMetadata" without a `name` field like "ExtendedCapabilities". I can see that you have handled that case in `wfs200.py` . 

[Here is an example with extended capabilities](https://api.dataforsyningen.dk/dhm_bluespot_ekstremregn?token=f3ecb52320902f733a433aa9945d8dc8&service=WMTS&request=GetCapabilities)

Since an early return is done in `OperationsMetadata`, the object is still created, but with invalid OperationMetadata object. This makes it so that you are not able to make the `gettile` request. It would maybe make more sence to have an error raised in `OperationsMetadata` constructor in this case as it is not valid anyways as the comment says. Then it might be easier to pinpoint the issue.
For the case in the `wmts` file, either check for the "ExtendedCapabilities" path or only include metadata starting with "Operation".

I do not fully understand the wmts standard so I do not know how often this will be the case or if there are other possible metadata that can be found inside OperationsMetadata.

```python
# ows.py, line 201
class OperationsMetadata(object):
    """Initialize an OWS OperationMetadata construct"""
    def __init__(self, elem, namespace=DEFAULT_OWS_NAMESPACE):
        if 'name' not in elem.attrib:  # This is not a valid element
            return
        self.name = elem.attrib['name']
        self.formatOptions = ['text/xml']
        parameters = []
        self.methods = []
        self.constraints = []
        ...

  def __repr__(self):
        return '<owslib.ows.OperationsMetadata {} at {}>'.format(self.name, hex(id(self)))
```

```python
# wmts.py, line 217
        ...
        self.operations = []
        serviceop = self._capabilities.find(_OPERATIONS_METADATA_TAG)
        #  REST only WMTS does not have any Operations
        if serviceop is not None:
            for elem in serviceop[:]:
                self.operations.append(OperationsMetadata(elem))
```

```python
# features/wfs200.py, line 122
        for elem in self._capabilities.find(nspath("OperationsMetadata"))[:]:
            if elem.tag != nspath("ExtendedCapabilities"):
                self.operations.append(OperationsMetadata(elem))
```

geographika:
Closed with #968.

PR #968, opened 2025-01-22
Filter out ExtendedCapabilities when reading WebMapTileService

geographika:
Fix for #965. 

Tested with the following:

```python
from owslib.wmts import WebMapTileService
service_url = "https://api.dataforsyningen.dk/dhm_bluespot_ekstremregn?token=f3ecb52320902f733a433aa9945d8dc8&"
wmts = WebMapTileService(service_url)

tile = wmts.gettile(layer='bluespot_ekstremregn_0',
                    tilematrixset='View1', tilematrix='L00',
                    row=0, column=0, format="image/png")
print(tile.info()['Content-Type'])

# without fix - AttributeError: 'OperationsMetadata' object has no attribute 'name'
```

CC @CJGutz - can you check if this fixes it for you. 

I'm considering adding a "live" test for this pointing to https://api.dataforsyningen.dk but endless test failures due to service changes are a concern. Long-term it may be better to spin up a local test server and return local XML and other files.

### DIFF
--- a/owslib/wmts.py
+++ b/owslib/wmts.py
@@ -34,7 +34,7 @@ import warnings
 from urllib.parse import (urlencode, urlparse, urlunparse, parse_qs,
                           ParseResult)
 from .etree import etree
-from .util import clean_ows_url, testXMLValue, getXMLInteger, Authentication, openURL, getXMLTree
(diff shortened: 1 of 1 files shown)


CJGutz (approved):
Yes, this works for me. Thank you!

PR #969, opened 2025-01-23
Minor README fixes

geographika:
- Fix reference to wrong library
- Remove `--universal` from release notes, so wheels don't contain reference to py2 (`OWSLib-0.32.1-py3-none-any.whl` instead of `OWSLib-0.32.1-py2.py3-none-any.whl`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ numerous OGC Web Service interfaces.
 
 ## Installation
 
-The easiest way to install pywis-pubsub is via the Python [pip](https://pip.pypa.io)
+The easiest way to install OWSLib is via the Python [pip](https://pip.pypa.io)
 utility:
 
 ```bash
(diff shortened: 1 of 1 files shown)


geographika:
Just running through the examples on the README.md and the server http://wms.jpl.nasa.gov/wms.cgi used in the examples is unfortunately no longer online.

The NEO Web Mapping Service (WMS) will also be taken offline in "early 2025" https://neo.gsfc.nasa.gov/blog/2024/08/30/neo-wms-web-mapping-service-deprecation-alert/ 

Any suggestions of WMS servers to use for the example?

requirements-dev.txt
```diff
@@ -9,3 +9,4 @@ twine
 wheel
 coverage
 coveralls
+build
```
tomkralidis, line 12: Do we still need `wheel`?

geographika: Removed

PR #971, opened 2025-02-15
OGC API: add support for collection level transactions

tomkralidis:
Note that tests are not implemented due to lack of public OGC APIs with this capability enabled.

### DIFF
--- a/owslib/ogcapi/features.py
+++ b/owslib/ogcapi/features.py
@@ -53,6 +53,57 @@ class Features(Collections):
         path = f'collections/{collection_id}/queryables'
         return self._request(path=path)
 
+    def collection_create(self, data: str) -> bool:
+        """
+        implements POST /collections
+
+        @type collection_id: string
+        @type data: string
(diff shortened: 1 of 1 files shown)


tomkralidis:
CI failure is intermittent.

kalxas (approved):

geographika (approved):

geographika:
@tomkralidis - yes, the intermittent failures relying on external services are a pain point in the project. Might be worth considering making the integration tests only run on demand rather than pull requests and/or consider a local test server returning the same content as the external services using something like https://pytest-httpserver.readthedocs.io

tomkralidis:
> @tomkralidis - yes, the intermittent failures relying on external services are a pain point in the project. Might be worth considering making the integration tests only run on demand rather than pull requests and/or consider a local test server returning the same content as the external services using something like https://pytest-httpserver.readthedocs.io

Neat idea -- can you do a (very small) proof of concept?

Issue #976, opened 2025-02-27
KeyError: 'content-type'


Xpunk:
Hi, there is an issue with 'content-type' not well managed:

To reproduce the issue based on this IGN url to get capabalities : 
https://data.geopf.fr/wfs/wfs?SERVICE=WFS&VERSION=2.0.0&REQUEST=GETCAPABILITIES

# Script :
from owslib.wfs import WebFeatureService
wfs = WebFeatureService(url='https://data.geopf.fr/wfs/wfs', version='2.0.0')
schema = wfs.get_schema('BDTOPO_V3:piste_d_aerodrome')
print(schema['geometry_column'])

# Env + libs used : 
python 3.11.9
OWSLib  0.32.0 / 0.32.1
requests 2.32.3

# Error :
Traceback (most recent call last):
  File "my_project_path\test_owslib.py", line 2, in <module>
    wfs = WebFeatureService(url='https://data.geopf.fr/wfs/wfs', version='2.0.0')
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "my_project_path\\.venv\Lib\site-packages\owslib\wfs.py", line 56, in WebFeatureService
    return wfs200.WebFeatureService_2_0_0(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "my_project_path\\.venv\Lib\site-packages\owslib\feature\wfs200.py", line 77, in __new__
    obj.__init__(
  File "my_project_path\\.venv\Lib\site-packages\owslib\feature\wfs200.py", line 128, in __init__
    self._capabilities = reader.read(self.url, self.timeout)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "my_project_path\\.venv\Lib\site-packages\owslib\feature\common.py", line 55, in read
    return getXMLTree(u)
           ^^^^^^^^^^^^^
  File "my_project_path\\.venv\Lib\site-packages\owslib\util.py", line 345, in getXMLTree
    content_type = rsp.info()['Content-Type']
                   ~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "my_project_path\\.venv\Lib\site-packages\requests\structures.py", line 52, in __getitem__
    return self._store[key.lower()][1]
           ~~~~~~~~~~~^^^^^^^^^^^^^
KeyError: 'content-type'

[...]

PR #977, opened 2025-02-28
Default to text/xml if server doesn't return a Content-Type

geographika:
Fix for #976. Assume content is `text/xml` if the server doesn't return a `Content-Type` header, as in the case of https://data.geopf.fr/wfs/wfs?service=WFS&request=GetCapabilities&version=2.0.0 which returns valid XML.


pvgenuchten:
Makes sense... thanx

Issue #973, opened 2025-02-25
NoneType object has no attribute attrib


pvgenuchten:
when querying the [csw endpoint of inspire](https://inspire-geoportal.ec.europa.eu/srv/eng/csw) on datasets of atmospheric conditions, i get this error

```
  File "/lib/python3.10/site-packages/owslib/iso.py", line 195, in __init__
    util.nspath_eval('gmd:characterEncoding/gmd:MD_CharacterSetCode', namespaces)).attrib.get(
AttributeError: 'NoneType' object has no attribute 'attrib'
```

not sure which record caused the error

geographika:
@pvgenuchten - are you able to provide the query that raised this error?

```python
url = 'https://inspire-geoportal.ec.europa.eu/srv/eng/csw'
c = cs(url)
c.getrecords2()
print(c.results)
# {'matches': 346714, 'returned': 10, 'nextrecord': 11}
# query with a filter?
# uuid_filter = fes.PropertyIsEqualTo(propertyname='sys.siteuuid', literal="{%s}" % "1706F520-2647-4A33-B7BF-592FAFDE4B45")
# c.getrecords2([uuid_filter], esn='full', maxrecords=2)
```

pvgenuchten:
code is this one:

``` 
from owslib.csw import CatalogueServiceWeb
from owslib.fes import PropertyIsEqualTo
url = 'https://inspire-geoportal.ec.europa.eu/srv/eng/csw'
csw = CatalogueServiceWeb(url)
cq = PropertyIsEqualTo('th_httpinspireeceuropaeutheme-theme.link', 'http://inspire.ec.europa.eu/theme/ac')
csw.getrecords2(constraints=[cq], outputschema='http://www.isotc211.org/2005/gmd', maxrecords=50, esn='full')     
``` 

indeed, code snippet is required to reproduce

geographika:
Closed with #978

PR #978, opened 2025-03-01
add a null check before using attrib

pvgenuchten:
resolves issue #973

### DIFF
--- a/owslib/iso.py
+++ b/owslib/iso.py
@@ -224,11 +224,14 @@ class PT_Locale(object):
             self.charset = None
         else:
             self.id = md.attrib.get('id')
-            self.languagecode = md.find(
-                util.nspath_eval('gmd:languageCode/gmd:LanguageCode', namespaces)).attrib.get('codeListValue')
-            self.charset = md.find(
-                util.nspath_eval('gmd:characterEncoding/gmd:MD_CharacterSetCode', namespaces)).attrib.get(
-                'codeListValue')
+            self.languagecode = None
+            languagecode = md.find(util.nspath_eval('gmd:languageCode/gmd:LanguageCode', namespaces))
+            if languagecode is not None:
+                self.languagecode = languagecode.attrib.get('codeListValue')
+            self.charset = None
+            charset = md.find(util.nspath_eval('gmd:characterEncoding/gmd:MD_CharacterSetCode', namespaces))
+            if charset is not None:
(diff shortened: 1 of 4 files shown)


geographika:
@pvgenuchten - I've added a test for this update based on the XML responses from the live server. Tests fail without your update, and pass with it. The PR was previously failing due to an unrelated reason (fixed in #982). 

@tomkralidis - as discussed at https://github.com/geopython/OWSLib/pull/971#issuecomment-2676181077 the test uses https://pytest-httpserver.readthedocs.io/ which allows request to be made by OWSLib, but using a mock server that can be configured to return XML from a saved file. It should be possible to run these in offline mode too (allowing only access to localhost) with a wrapper function in a conftest.py file - this can be addressed in a future PR.  

There is already a similar approach in OWSLib of loading XML from a file, for example: https://github.com/geopython/OWSLib/blob/39ad5a90c68925854424412a413b03728d38fcfa/tests/test_iso3_parsing.py#L302 

The advantage of https://pytest-httpserver.readthedocs.io/ is that it allows for testing of classes that don't allow XML to be passed in directly, and allows for end-to-end testing. 

Let me know your thoughts.

tomkralidis:
> @pvgenuchten - I've added a test for this update based on the XML responses from the live server. Tests fail without your update, and pass with it. The PR was previously failing due to an unrelated reason (fixed in #982).
> 
> @tomkralidis - as discussed at [#971 (comment)](https://github.com/geopython/OWSLib/pull/971#issuecomment-2676181077) the test uses https://pytest-httpserver.readthedocs.io/ which allows request to be made by OWSLib, but using a mock server that can be configured to return XML from a saved file. It should be possible to run these in offline mode too (allowing only access to localhost) with a wrapper function in a conftest.py file - this can be addressed in a future PR.
> 
> There is already a similar approach in OWSLib of loading XML from a file, for example:
> 
> https://github.com/geopython/OWS.Lib/blob/39ad5a90c68925854424412a413b03728d38fcfa/tests/test_iso3_parsing.py#L302
> 
> The advantage of https://pytest-httpserver.readthedocs.io/ is that it allows for testing of classes that don't allow XML to be passed in directly, and allows for end-to-end testing.
> 
> Let me know your thoughts.

Thanks for the info @geographika.  I think this would be a valuable update to our testing.  Feel free to open a dedicated issue so we can plan (and maybe split up the work if needed).

Issue #983, opened 2025-03-18
`owslib.wms.WebMapService` and `owslib.wfs.WebFeatureService` creation fails with `application/octet-stream` content-type services


SimonSAMPERE:
Hello,
I'm developing a plugin for QGIS and I've noticed that my code can no longer handle a particular OGC service in the latest stable version of QGIS. Here are the URLs of the services involved:
* WFS : https://inpn.mnhn.fr/webgeoservice/WFS/fxx_inpn
* WMS : https://inpn.mnhn.fr/webgeoservice/WMS/fxx_inpn

I've tried to investigate and to sum up, the following python code works well with `owslib` version 0.29.2 but not with version 0.32.0:

```python
from owslib.wms import WebMapService
from owslib.wfs import WebFeatureService


wfs_url = "https://inpn.mnhn.fr/webgeoservice/WFS/fxx_inpn?"
wfs_version = "2.0.0"

wfs = WebFeatureService(
    url= wfs_url,
    version = wfs_version
)
print(wfs.identification)

wms_url = "https://inpn.mnhn.fr/webgeoservice/WMS/fxx_inpn?"
wms_version = "1.3.0"

wms = WebMapService(
    url= wms_url,
    version = wms_version
)
print(wms.identification)
```

With `owslib` 0.32.0 a `ValuError` is raised by the `getXMLTree` method of `util.py`.

Here are some files with the entire error using owslib 0.32.0:
* for WFS: [wfs_error_owslib0320.txt](https://github.com/user-attachments/files/19318589/wfs_error_owslib0320.txt)
* for WMS: [wms_error_owslib0320.txt](https://github.com/user-attachments/files/19318587/wms_error_owslib0320.txt)

Perhaps the problem lies in the fact that these are services whose Content-Type is ‘application/octet-stream’ and not ‘application/xml’.
I don't know if this is a bug or if I just couldn't find the information on how to handle this with owslib 0.32.0. In any case, is there a workaround for this problem?

SimonSAMPERE:
I've implemented a workaround in the code of my plugin to parse the XML directly.
I hope that [this PR](https://github.com/geopython/OWSLib/pull/984) will be approved, that it will lead to the publication of a new version of `owslib` and that this new version will be quickly integrated into a future version of QGIS. 🤞🏼 
Anyway, congratulations and thank you @geographika for your responsiveness.

geographika:
[...]

PR #984, opened 2025-03-18
Test for XML in response, regardless of mimetype

geographika:
Fix for #983. `application/octet-stream` is a very generic mime type which could contain anything, so rather than adding more mime types to the list, a check is made to see if the response contains an xml tag, or is one of the following mime types:

```python
['text/xml', 'application/xml', 'application/vnd.ogc.wms_xml']
```

@SimonSAMPERE - can you check if this works with the WFS/WMS services that previously were failing?


SimonSAMPERE:
@geographika it works, well done!

Issue #987, opened 2025-04-11
[WMS] missing name warning in style parsing


pvgenuchten:
for some wms's i notice a large number of this warning

![Image](https://github.com/user-attachments/assets/eb4a899f-d465-4d0d-b4a3-d753c4028085)

it seems related to the style not having a name or title

is it an option to prevent the error after it's first occurrence

pvgenuchten:
yes, thank you @geographika

PR #990, opened 2025-04-14
do not show warnings on missing style name&title

pvgenuchten:
The warning about missing name|title in wms-style clogs our logs, because it is shown on each layer of a wms (if that wms has either key not set). On the other I don't see much value of the warning.

issue improves #987


geographika:
This makes sense to me. If both are missing then a warning is still emitted, otherwise title reverts to name if missing.

pvgenuchten:
my suggestion would be to merge this, and add new PR for WFS url update

PR #995, opened 2025-05-14
Add option to pass cookies to WMTS

ashergg:
Added an option to pass cookies to the WMTS class. This can help when the server requires specific cookies.

### DIFF
--- a/owslib/wmts.py
+++ b/owslib/wmts.py
@@ -136,7 +136,7 @@ class WebMapTileService(object):
             raise KeyError("No content named %s" % name)
 
     def __init__(self, url, version='1.0.0', xml=None, username=None, password=None,
-                 parse_remote_metadata=False, vendor_kwargs=None, headers=None, auth=None,
+                 parse_remote_metadata=False, vendor_kwargs=None, headers=None, cookies=None, auth=None,
                  timeout=30):
         """Initialize.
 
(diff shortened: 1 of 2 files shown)


owslib/wmts.py
```diff
@@ -136,7 +136,7 @@ def __getitem__(self, name):
             raise KeyError("No content named %s" % name)
 
     def __init__(self, url, version='1.0.0', xml=None, username=None, password=None,
-                 parse_remote_metadata=False, vendor_kwargs=None, headers=None, auth=None,
+                 parse_remote_metadata=False, vendor_kwargs=None, headers=None, cookies=None, auth=None,
```
geographika, line 139: It would be safer to add the new cookies parameter at the end of the list to avoid breaking existing codebases

owslib/wmts.py
```diff
@@ -887,7 +888,7 @@ class WMTSCapabilitiesReader:
     """Read and parse capabilities document into a lxml.etree infoset
     """
 
-    def __init__(self, version='1.0.0', url=None, un=None, pw=None, headers=None, auth=None):
+    def __init__(self, version='1.0.0', url=None, un=None, pw=None, headers=None, cookies=None, auth=None):
```
geographika, line 891: As above - move new argument to the end

geographika (changes requested):
Thanks @ashergg. A couple of comments to address. Ideall0,y a test would be added too.

geographika:
Thanks @ashergg ! This looks good to me, and +1 on merging. I'll leave open a few days to see if anyone else has any comments.

geographika (approved):

PR #996, opened 2025-05-14
Update WFS usage docs

KaiVolland:
This updates the docs for `Filter` usage with WFS.

I think that fes 2.0 is supported but needs a slightly different implementaton so i updated the docs.

You might want to check this, but this is how its working for me with WFS 2.0.2 and FES2


geographika:
It looks like the server used in the example is no longer online: http://geoserv.weichand.de:8080/geoserver/wfs

https://www.weichand.de/
> Die DemoServer (geoserv.weichand.de) stehen bis auf Weiteres nicht zur Verfügung.

I realise this server was used for the previous example, but if you are aware of another one we could use for the docs, and test this, that would be great.

KaiVolland:
We could use this one https://sgx.geodatenzentrum.de/wfs_vg250?REQUEST=GetCapabilities&SERVICE=WFS but im not sure if I'll find the time to update the whole WFS section.

pvgenuchten:
my suggestion would be to merge this, then create PR for url change

PR #1003, opened 2025-06-06
OGC version updates

geographika:
From @tomkralidis: 

> EDR is 1.0/1.1, not draft.  Records is also 1.0 now (press release coming soon from OGC). 

Question - the features.rst page lists versions of the standards supported by OWSLib (as well as latest versions of these standards)?


geographika:
As discussed - versions listed are versions supported by OWSLib.

PR #1004, opened 2025-06-09
ISO: add support for gmd:illuminationElevationAngle

tomkralidis:
- https://github.com/ISO-TC211/XML/blob/master/schemas.isotc211.org/19139/-/gmd/1.0/content.xsd#L81-L82
- https://github.com/ISO-TC211/XML/blob/master/schemas.isotc211.org/19115/-3/mrc/1.0/content.xsd#L199C40-L204
- https://github.com/ISO-TC211/XML/blob/master/schemas.isotc211.org/19115/-3/mrc/2.0/content.xsd#L202-L207


tomkralidis:
CI failures are unrelated/intermittent.

kalxas (approved):

PR #1006, opened 2025-06-15
OGC API: add support for sorting

tomkralidis:
Adds support for sorting in OGC API - Features/OGC API - Records.

### DIFF
--- a/owslib/ogcapi/features.py
+++ b/owslib/ogcapi/features.py
@@ -143,7 +143,7 @@ class Features(Collections):
                     sortby = sort_property
                 kwargs['sortby'] = sortby
             except ValueError as err:
-                LOGGER.debug(f'Cannot handle sortby; skipping')
+                LOGGER.debug(f'Cannot handle sortby; skipping: {err}')
 
         if 'cql' in kwargs:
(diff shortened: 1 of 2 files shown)


owslib/ogcapi/features.py
```diff
@@ -124,14 +124,23 @@ def collection_items(self, collection_id: str, **kwargs: dict) -> dict:
...
         @returns: feature results
         """
 
         if 'bbox' in kwargs:
             kwargs['bbox'] = ','.join(list(map(str, kwargs['bbox'])))
         if 'datetime_' in kwargs:
-            kwargs['datetime'] = kwargs['datetime_']
+            kwargs['datetime'] = kwargs.pop('datetime_')
+        if 'sortby' in kwargs:
+            sort_property, sort_order = kwargs.pop('sortby')
```
geographika, line 138: Is there a risk of `ValueError: too many values to unpack (expected 2)` if a querystring is missing a value? `sortby=prop1`. Could a default be set for the sort order?

geographika (approved):
Is it possible to add a `sortby` example to the usage docs (or a live test)?

PR #1012, opened 2025-09-17
WCS: ensure identifier is a string (not a list) (#1010)

### LINKED ISSUE #1010: Passing an identifier string (not in a list) silently truncates it
Hello,

I've got bitten by this bug (not sure if it is fair to call it a bug, maybe more a footgun):

```
        assert len(identifier) > 0
        request["CoverageID"] = identifier[0]
```
The code above that checks for the coverage will silently truncate the coverage id if it is passed as a single string. i.e.

```
    response = wcs.getCoverage(
        identifier="some_id",
        format="image/tiff",
        subsets=[
            ("lat", bounds.south, bounds.north),
            ("lon", bounds.west, bounds.east)
        ],
        interpolation="nearest"
    )
```

If the desired api is using only lists, then It would make more sense to check for it directly and possibly error early explaining what happened.

Thanks for your work!

tomkralidis:
Fixes #1010

### DIFF
--- a/owslib/coverage/wcs100.py
+++ b/owslib/coverage/wcs100.py
@@ -102,7 +102,7 @@ class WebCoverageService_1_0_0(WCSBase):
         note: additional **kwargs helps with multi-version implementation
         core keyword arguments should be supported cross version
         example:
-        cvg=wcs.getCoverage(identifier=['TuMYrRQ4'], timeSequence=['2792-06-01T00:00:00.0'], bbox=(-112,36,-106,41),
+        cvg=wcs.getCoverage(identifier='TuMYrRQ4', timeSequence=['2792-06-01T00:00:00.0'], bbox=(-112,36,-106,41),
                             format='cf-netcdf')
 
         is equivalent to:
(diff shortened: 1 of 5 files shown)


tomkralidis:
FYI CI failures are unrelated to this PR.

geographika:
Thanks @tomkralidis.

Looks like CI is failing due to (unrelated) http://34.67.197.57:8585 being offline. Used in https://github.com/geopython/OWSLib/blob/7c4ca6ff3c2ab804307c4c561705a2334f7f6447/tests/test_ogcapi_connectedsystems_osh.py#L22-L23

Was using a list for the identifier an error (I presume so as it is no longer supported)?

tomkralidis:
> Was using a list for the identifier an error (I presume so as it is no longer supported)?

It was for downstream applications passing `identifier` as a `str` in `getCoverage`.  The PR removes the ambiguity.

geographika:
> > Was using a list for the identifier an error (I presume so as it is no longer supported)?
> 
> It was for downstream applications passing `identifier` as a `str` in `getCoverage`. The PR removes the ambiguity.

Ok. +1 for merging, but as this will break existing user code, I guess we should bump the version to `0.36.0`.

tomkralidis:
> Ok. +1 for merging, but as this will break existing user code, I guess we should bump the version to 0.36.0.

Yes, +1 on next release.

PR #1014, opened 2025-10-18
increase ISO contact parsing

tomkralidis:


tomkralidis:
> contributor

`owslib.iso.MD_DataIdentification` and `owslib.iso3.MD_DataIdentification` have `creator`, `publisher`, `contributor` members (lists) (note that ISO3 additionally includes `funder`).  These are populated on parsing if they exist (as `CI_RoleCode` values).

In summary, `owslib.iso` and `owslib.iso3` `get_all_contacts` is convenience function which attempts to capture all contacts in an ISO document.

PR #1016, opened 2025-10-18
WMS 1.3.0: support any boolean for queryable layers (#993)

### LINKED ISSUE #993: WMS 1.3.0 should allow boolean values in WMS-Capabilities
Hi there,

Since WMS 1.3.0, some attributes of WMS Layer accept boolean values (`true`, `false`) and not only integer values (`0`,`1`). See https://portal.ogc.org/files/?artifact_id=14416, page 30.

Today, OWSLib is only allowing integer values, for example for the `queryable` attribute: https://github.com/geopython/OWSLib/blob/master/owslib/map/wms130.py#L449

I think boolean values should also be accepted for WMS 1.3.0 here. 
What do you think?
Thanks,
Guillaume.

tomkralidis:
Fixes #993 (as well as some code housekeeping).

### DIFF
--- a/owslib/map/wms130.py
+++ b/owslib/map/wms130.py
@@ -2,9 +2,11 @@
 # =============================================================================
 # Copyright (c) 2004, 2006 Sean C. Gillies
 # Copyright (c) 2005 Nuxeo SARL <http://nuxeo.com>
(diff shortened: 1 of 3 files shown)


geographika:
Thanks @tomkralidis !

PR #1020, opened 2025-10-20
Fix str2bool call

geographika:


geographika:
@tomkralidis - we'd have caught this if the test suite was passing! CI should now all be green again with this change. Thoughts on a new release later this week?

PR #1025, opened 2025-11-04
remove comments from xml, because it may fail parsing it

pvgenuchten:
found an issue in a xml with a comment before <srv:service-identification>
suggestion to remove any comments before parsing

also added a new test case, pre (fails) and post (succeeds) PR


pvgenuchten:
Not sure if comments are used by owslib anywhere, removed all of them in case similar code is used in other places causing this error

PR #1028, opened 2025-11-30
docs: update WMS example

tomkralidis:
This PR updates the WMS example in the documentation against a working WMS instance.


tomkralidis:
CI failures are unrelated.

geographika:
Thanks @tomkralidis!
Re the failing test, I'll try and capture the output when (if) http://nsidc.org/cgi-bin/atlas_south?service=WFS&request=GetCapabilities&version=1.0 comes back online and switch to a local test. 

It might be a nice use of GeoHealthCheck to monitor any of the live sites used in the test suite.

PR #1031, opened 2026-02-28
use only pyproject configuration

tomkralidis:
This PR:

- moves to using `pyproject.toml` exclusively for project configuration
- removes `setup.py` and related artifacts
- note that in `pyproject.toml` we pin `setuptools` to `<69` for Ubuntu 24.04 compatibility

This PR also contains some test updates (note that the changes in `tests/test_wfs_generic.py` are due to the remote WFS service error'ing out on DescribeFeatureType requests only in WFS 1.0.0 mode; test is removed given we support WFS 1.1.0 and 2.0.0 proper).


geographika:
@tomkralidis - I checked out this branch and built locally. All runs fine, but the wheel is missing all the subpackages of owslib. Image on the left is the new wheel, on the right the latest on pypi:

image

I tried adding the following, but for some reason the `tests` folder still gets added to the `sdist` `.tar.gz` (which didn't include it before). 

```toml
#[tool.setuptools]
#packages = ["owslib"]

# find all subpackages automatically
[tool.setuptools.packages.find]
where = ["owslib"]
```

Some related long discussions at https://github.com/pypa/setuptools/discussions/3346 - maybe this is fixed in newer setuptools? Have you run into it before?

tomkralidis:
@geographika it turns out that `MANIFEST.in` helps here (removing `tests/`).  Let me know if `dist/` is now fixed on build.

geographika (approved):

cehbrecht (approved):

landryb:
fwiw, pining setuptools to `<69` "for ubuntu compat" is a bit sad for the rest of the downstream ecosystem, because we have to patch it out to package on OpenBSD - where owslib 0.36 packages just fine with setuptools 80.9.

tomkralidis:
> fwiw, pining setuptools to `<69` "for ubuntu compat" is a bit sad for the rest of the downstream ecosystem, because we have to patch it out to package on OpenBSD - where owslib 0.36 packages just fine with setuptools 80.9.

See #1041 (thanks @geographika) for PR/update.  @landryb let us know if that helps on OpenBSD?

landryb:
thanks, i had it [locally patched out](https://github.com/openbsd/ports/blob/f65fcfd4dbae09446fd0189cc1589870d8cc4d6e/geo/py-owslib/patches/patch-pyproject_toml), `>=77` should be fine too

PR #1032, opened 2026-02-28
Update contents of wheel and sdist

geographika:
@tomkralidis - these changes are what work for me when building #1031. 

Only `tests` needed to be excluded in MANIFEST.in as it is the only folder with a `__init__.py` file. 

`include = ["owslib*"]` was the only way I could get all the OWSLib files added to the wheel. I can't say I fully understand why `where = ["owslib"]` didn't work here. 

Maybe check what you get with a build of this PR and `python -m build`?

(note the PR is for your `pyproject` branch)

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,2 @@
-prune binder
-prune etc
+# this file controls the contents of sdist
 prune tests
-prune examples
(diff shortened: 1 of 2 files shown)


tomkralidis:
@geographika thanks, merged.  #1031 ready for review again.

Issue #865, opened 2023-03-01
OGC API bug ignores SSL verify flag of Authentication object


berntholer:
Our corporate firewall inserts its own self signed certificate in the SSL chain. For quick development I usually test things using the `verify=False` option in request calls.

When I tested some WMS functionality of OWSlib I first created a simple Authentication object like so: 
```
myauth=Authentication(verify=false)
```
Passing `myauth` to WebMapService(..., auth=myauth) worked as expected. But when I tried to do the same for OGC API Features(..., auth=myauth) calls I got SSL self signed certificate errors.

Upon inspecting the code I saw that the auth.verify value is never used in the `http_prepare()` function, unlike in the `openURL()` function used for i.e. WMS.

I created a PR to fix this 37cfb31ef8fbcb81efc7020497c3b9cf39bd8592

PS I assume another potential problem lies with the pop() statements like [this ](https://github.com/berntholer/OWSLib/blob/e8dd55fb1d349dde0e2c7881fa9103681708201c/owslib/util.py#L462) followed by a get() later on [here](https://github.com/berntholer/OWSLib/blob/e8dd55fb1d349dde0e2c7881fa9103681708201c/owslib/util.py#L473) which will always be empty since the value was popped out of the dictionary?

PR #1040, opened 2026-06-07
fix http_prepare handling of cert/verify

tomkralidis:
This PR fixes `owslib.util:http_prepare` to preserve cert and verify parameters as defined by the caller.

Note this fix is in support of https://github.com/qgis/QGIS/issues/66296.


tomkralidis:
CI failures are unrelated.

agiudiceandrea:
Does this fix https://github.com/geopython/OWSLib/issues/865?

tomkralidis:
> Does this fix #865?

It should, yes.

tests/test_ogcapi_records_pycsw.py
geographika: We could just add a > 0 here to avoid having to keep in sync with the pycsw changes?

geographika (approved):

tests/test_ogcapi_records_pycsw.py
tomkralidis: > We could just add a > 0 here to avoid having to keep in sync with the pycsw changes?

Good suggestion; done.

PR #1041, opened 2026-06-14
Remove setuptools constraint and test builds

geographika:
See comment from @landryb at https://github.com/geopython/OWSLib/pull/1031#issuecomment-4701033461

This PR removes the constraint and tests the wheel builds in the CI. 

@tomkralidis - is the setuptools constraint still valid?

Also fix:

```
Please use a simple string containing a SPDX expression for project.license. You can also use project.license-files. (Both options available on setuptools>=77.0.0).

 By 2027-Feb-18, you need to update your project and remove deprecated calls
 or your builds will no longer be supported.
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,6 +22,8 @@ jobs:
         pip3 install ".[dev]"
         pip3 install ".[docs]"
         pip3 install ".[test]"
+    - name: run flake8 ⚙️
+      run: flake8 owslib/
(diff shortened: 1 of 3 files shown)


tomkralidis:
> @tomkralidis - is the setuptools constraint still valid?

Thanks for the PR @geographika, we'll have to test against noble based packaging, but fine to move ahead (where noble based packages can pin to an earlier OWSLib version).

